Issue #5, opened 2017-06-17
Enable APIs


jennybc:
I've gone here:

https://console.developers.google.com

logged in with the necessary Google identify and enabled the Google Drive and Sheets API for the gargle-associated project. I suspect more will need to be enabled in the long run, hence this issue. But I decided to do on an as needed basis. None of this matters until I adopt the new app credentials yet, but ~~I will do that soon~~ that's in #6.

@craigcitro Let me know if there are other APIs I should immediately enable.

craigcitro:
Only other ones I can think of immediately are BQ and GCS -- but IIRC both of those are enabled by default.

jennybc:
OK I will close. We can explicitly enable more if we discover it's necessary.

PR #6, opened 2017-06-17
Update gargle app

jennybc:
This is the gargle client id and secret now FYI. Only Drive and Sheets APIs currently enabled, see #5

### DIFF
--- a/R/registry.R
+++ b/R/registry.R
@@ -9,7 +9,7 @@
 #'
 #' @format An environment.
 #' @keywords internal
-gargle_env <- new.env()
+gargle_env <- new.env(parent = emptyenv())
 gargle_env$credential_functions <- list()
 
 #' Check that f is a viable credential fetching function.
--- a/R/three-legged-oauth.R
+++ b/R/three-legged-oauth.R
@@ -5,12 +5,16 @@
 #' @param ... Additional arguments (ignored)
 #' @export
 get_user_oauth2_credentials <- function(scopes, oauth_app = NULL, ...) {
-  
-  endpoint <- httr::oauth_endpoints("google")
-  app <- httr::oauth_app("google",
-                         "465736758727.apps.googleusercontent.com",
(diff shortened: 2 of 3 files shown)


R/registry.R
```diff
@@ -9,7 +9,7 @@
 #'
 #' @format An environment.
 #' @keywords internal
-gargle_env <- new.env()
+gargle_env <- new.env(parent = emptyenv())
```
craigcitro: Just for my own understanding: this is just good hygiene, to make sure we aren't accidentally picking up entries in the parent?

jennybc: Yeah. [To quote](http://adv-r.had.co.nz/Environments.html#explicit-envs) @hadley:

> Just as you can use a list to pass data between functions, you can also use an environment. When creating your own environment, note that you should set its parent environment to be the empty environment. This ensures you don’t accidentally inherit objects from somewhere else.

R/three-legged-oauth.R
```diff
@@ -5,12 +5,16 @@
...
+  app <- httr::oauth_app(
+    "google",
+    key = "603366585132-nku3fbd298ma3925l12o2hq0cc1v8u11.apps.googleusercontent.com",
+    secret = "as_N12yfWLRL9RMz5nVpgCZt"
+  )
+  token <- httr::oauth2.0_token(
+    endpoint = httr::oauth_endpoints("google"),
+    app = app,
+    scope = scopes
+  )
```
hadley: Did you deliberately drop `use_oob = FALSE`?

jennybc: No. Now restored 4e8e1b763e7c27160992f09e405c2aacef7e8e60.

craigcitro: FWIW, I think it might have been there for debugging purposes; we probably want to just pick up `httr`'s default, right?

(I pondered this during the initial review and just decided that @jennybc was one step ahead of me on realizing we wanted the default.)

PR #7, opened 2017-06-18
Use build matrix on travis

jennybc:
Does this build matrix look ok? It's a simplified version of one I've used elsewhere but I confess to not knowing exactly when I need to explicitly specify linux trusty vs. precise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,23 @@
 language: r
+cache: packages
 
-r_github_packages:
-  - jimhester/covr
+matrix:
+  include:
+    - os: linux
+      r: oldrel
+    - os: linux
+      dist: trusty
+      env: R_CODECOV=true
+      r: release
+    - os: linux
+      r: devel
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,7 +1,23 @@
...
+cache: packages
 
-r_github_packages:
-  - jimhester/covr
+matrix:
+  include:
+    - os: linux
+      r: oldrel
+    - os: linux
+      dist: trusty
```
craigcitro: Does it work if we just hoist

    - dist: trusty
    - sudo: false

up to top-level? In particular, I don't think we need `sudo` for anything, meaning it's hopefully faster (or, at least, [that's what it says on the label](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments)).

jennybc: We shall find out: e0c6a1b7bb9ba0fd677d154aa23058bf5a6973d5

PR #15, opened 2017-07-21
Tidyverse & gargle apps, tweaks to three-legged oauth credfun

jennybc:
*Updated*

Move tidyverse app into its own function so we can be more clear about who should be using it and who should not (as previously discussed w/ @hadley).

Create a gargle app to use as default, for general tire-kicking.

~~Expose `cache` argument in `credentials_user_oauth2()`.~~ Pass `...` through to `httr::oauth2.0_token()`.

~~Closer alignment with~~ Inherit `httr::oauth2.0_token()` parameter names and docs.

`tidyverse_app()` in action:

``` r
devtools::load_all(".")
#> Loading gargle
tidyverse_app()
#> <oauth_app> tidyverse
#>   key:    603366585132-nku3fbd298ma3925l12o2hq0cc1v8u11.apps.googleusercontent.com
#>   secret: <hidden>
```

`token_fetch()` in action:

``` r
devtools::load_all(".")
#> Loading gargle
scope <- "https://www.googleapis.com/auth/drive"
(x <- token_fetch(scope, app = tidyverse_app()))
#> <Token>
#> <oauth_endpoint>
#>  authorize: https://accounts.google.com/o/oauth2/auth
#>  access:    https://accounts.google.com/o/oauth2/token
#>  validate:  https://www.googleapis.com/oauth2/v1/tokeninfo
#>  revoke:    https://accounts.google.com/o/oauth2/revoke
#> <oauth_app> tidyverse
#>   key:    603366585132-nku3fbd298ma3925l12o2hq0cc1v8u11.apps.googleusercontent.com
#>   secret: <hidden>
#> <credentials> access_token, expires_in, refresh_token, token_type
#> ---
```


R/three-legged-oauth.R
```diff
@@ -1,21 +1,35 @@
...
+#'   appname = "my_awesome_app",
+#'   key = "keykeykeykeykeykey",
+#'   secret = "secretsecretsecret"
+#' )
+#'
+#' ## credentials_user_oauth2() is one of the functions token_fetch() will try
+#' token_fetch(scope, app = app)
+#' }
+credentials_user_oauth2 <- function(scopes,
+                                    app,
```
craigcitro: Should we default to the tidyverse app?

Or default to another clearly-labeled-default app and let tidyverse libraries override? (I prefer this one.)

In particular, I think it's nice for users to be able to call `credentials_user_oauth2` by hand to kick the tires without first having to run the gauntlet of making/understanding client IDs. 😉

R/three-legged-oauth.R
```diff
@@ -1,21 +1,35 @@
...
+                                    cache = getOption("httr_oauth_cache"),
+                                    ...) {
   token <- httr::oauth2.0_token(
     endpoint = httr::oauth_endpoints("google"),
     app = app,
     scope = scopes,
-    use_oob = FALSE
+    use_oob = FALSE,
```
craigcitro: Not in this PR, but should we default this one to `TRUE`? (Maybe this says something about the oddball environments I tend to live in?)

NAMESPACE
```diff
@@ -11,5 +11,6 @@ export(credentials_gce)
 export(credentials_service_account)
 export(credentials_travis)
 export(credentials_user_oauth2)
+export(tidyverse_app)
```
craigcitro: Given that we don't intend for *most* users to use this directly, should we instead leave this unexported and let tidyverse packages do a `:::` to find it? (That seems really dirty, but am curious if there's an accepted R idiom for this sort of thing.)

jennybc: Sadly, the other tidyverse packages won't be able to pass `R CMD check` and get on CRAN if we do this. It would be nice, though.

R/three-legged-oauth.R
jennybc: This sounds like a great idea. Sure, I can create a default gargle app and we can say it's for tire-kicking vs. building into a package.

R/three-legged-oauth.R
jennybc: I was wondering why this isn't `getOption("httr_oob_default")` and/or why it's not exposed. If we actually consulted `...` both `use_oob` and `cache` would default to httr default, but be possible to override.

craigcitro: +1 on trying to pass this through to `httr`; I wasn't sure the cleanest way to do it.

In particular, passing `...` on to `oauth2.0_token` is nice, but had three issues:
* we can never add new args unless we know they don't collide with `oauth2.0_token` args
* bad times if we ever migrate off `oauth2.0_token`
* I didn't know how (or if it's possible to) get roxygen to inherit args from a function in another package for doc purposes.

I think a reasonable option here is probably to just have a `token_args` option, and anything in that list gets passed along to `oauth2.0_token`. This gives us a nice escape hatch in the event of `oauth2.0_token` being changed: we can deprecate that arg.

WDYT?

R/three-legged-oauth.R
jennybc: done in dd818c5b7953378f60c47e9d3f2bcc762416fa98

R/three-legged-oauth.R
jennybc: dbeb4963f0036d966f4d89a6c698c83fed3df892 does the simplest thing: passing `...` through and uses some nice new features in roxygen2 to finesse the docs.

R/gargle-oauth-app.R
```diff
@@ -0,0 +1,22 @@
+#' Oauth app for demonstration purposes
```
hadley: OAuth?

R/tidyverse-oauth-app.R
```diff
@@ -0,0 +1,20 @@
+#' Oauth app for tidyverse packages
```
hadley: OAuth

R/tidyverse-oauth-app.R
```diff
@@ -0,0 +1,20 @@
...
+#'
+#' Returns an OAuth app for use in tidyverse packages, e.g., googledrive,
+#' googlesheets, bigrquery. Please don't use this app directly in non-tidyverse
+#' projects.
+#'
+#' @seealso For a default app to use while getting to know gargle and
+#'   instructions on how to make your own app, see [gargle_app()].
+#'
+#' @return An OAuth consumer application, produced by [httr::oauth_app()].
+#' @export
```
hadley: `@keyword internal`

jennybc:
@hadley I addressed 'Oauth' --> 'OAuth' and the internal keyword for `tidyverse_app()`.

I think the only substantive thing is whether to pass `...` from `token_fetch()` to `httr::oauth2.0_token()` or to do something more specific, as @craigcitro raised. This is the first instance of a credential fetching method passing `...` along to another function vs. just taking what it needs, e.g. the path to the JSON for a service token or the app for OAuth2.

Shall I merge it?

R/three-legged-oauth.R
hadley: I think we can work on the assumption that if we ever use something other than `oauth2.0_token` it would still have to use the same args. I think passing `...` is fine.

PR #17, opened 2017-07-24
Auth state

jennybc:
This is WIP, to give something specific to talk about.

Main addition: gargle would maintain an auth state, which consists of:

  * active vs inactive:
    - active ➡️ a token request gets a token (including fetching one, if necessary)
    - inactive ➡️ a token request gets `NULL`
  * token: a token. Initially set to `NULL` and reset to `NULL` by `auth_deactivate(clear = TRUE)`
  * method: character, name of the credential fetching method that produced the token in `token`

Is this general idea ok? If we can agree on that, I'm sure there are refinements re: how it's done.

I want to be able to explicitly turn off auth so that, for example, you can download a world-readable file w/o being challenged for any credentials.

I think it's clear that someone, either gargle or the client package, has to hold on to the token for the duration of a "session" and it seems better to do in gargle. I can justify that more if necessary.

I have a way to ingest a token from an object or a token saved to file. This might be a weird vestigial thing I got used to doing in googlesheets and, therefore, it exists in googledrive. We could discuss whether that should be possible, but I need it for now.

I have a companion branch in googledrive that, together with this, allows me to basically remove all auth functionality from that package and still pass all tests. It works (crudely) with 3-legged oauth and a service account token already.

Other notes:

  * ~~the base is the branch from #15~~ that PR is now merged and this one is updated to reflect
  * there's a lot of chatty messages that are just temporary


jennybc:
Here's a little live usage, calling the gargle functions directly, even though I can do all of this through googledrive:

```r
> devtools::load_all(".")
Loading gargle
gargle onLoad!
> auth_state()
auth: ACTIVE
token state: NULL
last credential method: NA
> token_deliver("https://www.googleapis.com/auth/drive")
in token_fetch()
trying credentials_service account
[1] "Path must end in .json"
trying credentials_travis
trying credentials_gce
trying credentials_user_oauth2
Press Esc/Ctrl + C to abort
Authentication complete.
token fetched!
<Token>
<oauth_endpoint>
 authorize: https://accounts.google.com/o/oauth2/auth
 access:    https://accounts.google.com/o/oauth2/token
 validate:  https://www.googleapis.com/oauth2/v1/tokeninfo
 revoke:    https://accounts.google.com/o/oauth2/revoke
<oauth_app> gargle-demo
  key:    603366585132-orjlfqlnkvnkeb1menfdhss2oej4i2d5.apps.googleusercontent.com
  secret: <hidden>
<credentials> access_token, expires_in, refresh_token, token_type
---
> auth_state()
auth: ACTIVE
token state: CACHED
last credential method: user_oath2
> googledrive::drive_find(n_max = 2)
# A tibble: 2 x 3
                                                name
*                                              <chr>
1                                 UseR! 2017 Debrief
2 RStudio Cloud Notes from interviews with prospects
# ... with 2 more variables: id <chr>, files_resource <list>
> gargle::auth_deactivate()
auth: INACTIVE
token state: CACHED
last credential method: user_oath2
> googledrive::drive_find(n_max = 2)
Error: HTTP error [403] The user does not have sufficient permissions for this file.
  * domain: global
  * reason: insufficientFilePermissions
  * message: The user does not have sufficient permissions for this file. 
> gargle::auth_activate()
auth: ACTIVE
token state: CACHED
last credential method: user_oath2
> googledrive::drive_find("UseR!")
# A tibble: 1 x 3
                name                                           id files_resource
*              <chr>                                        <chr>         <list>
1 UseR! 2017 Debrief 1iVYtIdhG27ARdZXqBnY4iX5Vg4OHDmyKFCPKFnfux_g    <list [25]>
> gargle::auth_deactivate(clear = TRUE)
TO DO: should deactivate cache file right here
auth: INACTIVE
token state: NULL
last credential method: 
> googledrive::drive_find(n_max = 2)
Error: HTTP error [403] The user does not have sufficient permissions for this file.
  * domain: global
  * reason: insufficientFilePermissions
  * message: The user does not have sufficient permissions for this file. 
> token_fetch("https://www.googleapis.com/auth/drive", path = "gargle-e7c9c4883ee9.json")
in token_fetch()
trying credentials_service account
token fetched!
<Token>
NULL
NULL
<credentials> access_token, expires_in, token_type
---
> auth_state()
auth: ACTIVE
token state: CACHED
last credential method: service_acount
> googledrive::drive_find()
# A tibble: 2 x 3
             name                            id files_resource
*           <chr>                         <chr>         <list>
1     DESCRIPTION  0B6vwoYQ8r5jIR1g1UWEtOXkzR1U    <list [33]>
2 Getting started 0B6vwoYQ8r5jIc3RhcnRlcl9maWxl    <list [31]>
```

R/aaa.R
```diff
@@ -0,0 +1,26 @@
...
+#' * `$auth`: a list that records current auth state:
+#'   - `$active` Logical, indicates whether a token or `NULL` should be
+#'     delivered upon request via `token_deliver()`.
+#'   - `$token` The most recently fetched token. If none has been fetched, value
+#'     is `NULL`.
+#'   - `$method` The name of the credential function that produced the token. If
+#'     none has been used, value if `NA_character_`.
+#'
+#' @format An environment.
+#' @keywords internal
```
hadley: I think you should `@noRd` this since it's purely internal

jennybc: done dc6ac910a92ce11e2a9288179d7bcb2ecc36e50f

jennybc:
Question One is: do we agree gargle should maintain an auth state, including at least the current token but probably more?

If yes, then we should discuss the actual design. I just have something here that fills the particular hole left by ripping this functionality out of googledrive. And that design came from googlesheets, which might be bad and shouldn't be planned around.

craigcitro:
I am 👍 on the idea of having some explicit auth state by default. Happy to ponder details, but I suspect that if we had something that worked for all three of bigrquery/googledrive/googlestats, then it's awesome by definition. 😉 

I will look at why I didn't use the existing service account call right now and see if I remember what I was thinking.

craigcitro:
So a potentially relevant observation is that the service account JSON file retrieved from cloud console contains its own info about the OAuth endpoint; it's possible that it doesn't match the default Google endpoint in rare cases, but likely only for google-internal use (and not a showstopper).

I think we can move that to a separate issue/PR; in particular, if using the default google app Just Works, then I'd say let's do it.

jennybc:
The vague 👍 from @hadley and @craigcitro is enough for now. I will think on a more proper implementation. I think gargle should be checking (at least) the scopes and app and confirming a match before blithely emitting the current token. I might have to just take notes somewhere for discussion.

jennybc:
I now think such logic belongs in the client package, e.g., googledrive. Maybe after doing that for googledrive and, e.g., gmailr, it will be clear is some parts actually do belong in gargle? But for now I am backing out.

PR #30, opened 2017-11-20
Refactor cache

jennybc:
WIP to discuss in 1-on-1

  * gargle gets own options, `gargle.oob_default` and `gargle.oauth_cache`, instead of consulting those for httr. Same defaults, though.
  * All the (unexported) caching functions from httr come over, so they can gain new defaults or be modified. Allowed me to eliminate the `:::`s I was tolerating for a while.
  * Default cache path is `~/.R/gargle/gargle-oauth` as opposed to `.httr-oauth` in working directory.
  * Store oauth tokens in a tibble, as opposed to a list. Currently just a formal tweak, as the tibble only has two variables: the usual hash identifier and a list-column of Gargle2.0 objects.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -14,11 +14,12 @@ Description: Utilities to take some of the pain out of dealing with
     their users.
 License: MIT + file LICENSE
 LazyData: true
-Imports:
+Imports: 
     debugme,
     httr,
     jsonlite,
-    magrittr
+    magrittr,
+    tibble
 Suggests:
     testthat
 RoxygenNote: 6.0.1.9000
--- a/R/Gargle-class.R
+++ b/R/Gargle-class.R
@@ -22,10 +22,10 @@ gargle2.0_token <- function(email = NULL,
                             scope = NULL,
                             user_params = NULL,
                             type = NULL,
-                            use_oob = getOption("httr_oob_default"),
+                            use_oob = getOption("gargle.oob_default"),
                             ## params end
                             credentials = NULL,
-                            cache = getOption("httr_oauth_cache"), ...) {
+                            cache = getOption("gargle.oauth_cache"), ...) {
   params <- list(
     scope = scope,
     user_params = user_params,
@@ -67,7 +67,7 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
                         app = gargle_app(),
                         credentials = NULL,
                         params = list(),
-                        cache_path = getOption("httr_oauth_cache")) {
+                        cache_path = getOption("gargle.oauth_cache")) {
     "!DEBUG Gargle2.0 initialize"
     stopifnot(
       httr:::is.oauth_endpoint(endpoint) || !is.null(credentials),
--- a/R/Gargle-class.R
+++ b/R/Gargle-class.R
@@ -70,8 +70,8 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
                         cache_path = getOption("gargle.oauth_cache")) {
     "!DEBUG Gargle2.0 initialize"
     stopifnot(
-      httr:::is.oauth_endpoint(endpoint) || !is.null(credentials),
-      httr:::is.oauth_app(app),
+      is.oauth_endpoint(endpoint) || !is.null(credentials),
+      is.oauth_app(app),
       is.list(params)
     )
 
@@ -80,7 +80,7 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
     self$endpoint <- endpoint
     params$scope <- add_email_scope(params$scope)
     self$params <- params
-    self$cache_path <- httr:::use_cache(cache_path)
+    self$cache_path <- use_cache(cache_path)
 
     if (!is.null(credentials)) {
       # Use credentials created elsewhere - usually for tests
@@ -149,7 +149,7 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
       cached <- fetch_matching_tokens(self$hash(), self$cache_path)
     } else {
       "!DEBUG searching cache for matches on endpoint + app + scopes + email: `sQuote(self$email)`"
-      cached <- httr:::fetch_cached_token(self$hash(), self$cache_path)
+      cached <- fetch_cached_token(self$hash(), self$cache_path)
     }
 
     if (is.null(cached)) return(FALSE)
@@ -165,7 +165,8 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
 fetch_matching_tokens <- function(hash, cache_path) {
   if (is.null(cache_path)) return(NULL)
 
-  tokens <- httr:::load_cache(cache_path)
+  "!DEBUG `cache_path`"
+  tokens <- load_cache(cache_path)
   matches <- mask_email(names(tokens)) == mask_email(hash)
 
   if (!any(matches)) return(NULL)
--- a/R/Gargle-class.R
+++ b/R/Gargle-class.R
@@ -141,11 +141,17 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
     # append the email
     paste(msg, self$email, sep = "-")
   },
+  cache = function(path = self$cache_path) {
+    "!DEBUG cache a token"
+    cache_token(self, path)
+    self
+  },
   load_from_cache = function() {
     if (is.null(self$cache_path)) return(FALSE)
 
     if (is.null(self$email)) {
       "!DEBUG searching cache for matches on endpoint + app + scopes"
+      "!DEBUG cache_path is `sQuote(self$cache_path)`"
       cached <- fetch_matching_tokens(self$hash(), self$cache_path)
     } else {
       "!DEBUG searching cache for matches on endpoint + app + scopes + email: `sQuote(self$email)`"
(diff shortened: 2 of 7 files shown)


DESCRIPTION
```diff
@@ -14,11 +14,12 @@ Description: Utilities to take some of the pain out of dealing with
     their users.
 License: MIT + file LICENSE
 LazyData: true
-Imports:
+Imports:
```
jennybc: see?! why?

R/Gargle-class.R
```diff
@@ -141,15 +141,21 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
     # append the email
     paste(msg, self$email, sep = "-")
   },
+  cache = function(path = self$cache_path) {
+    "!DEBUG cache a token"
+    cache_token(self, path)
+    self
+  },
```
jennybc: This is same as that in `Token` class in httr, but seems needed in here so that `gargle::cache_token()` is used instead of `httr::cache_token()`.

R/oauth-cache.R
```diff
@@ -0,0 +1,138 @@
+## see oauth-cache.R in httr
+## nothing there is exported, so to make desired changes to oauth caching
+## I have to copy and mutate here
+
+gargle_cache_path <- function() {
+  "~/.R/gargle/gargle-oauth"
+}
```
jennybc: New default cache path ➡️ user-level storage is preferred.

R/oauth-cache.R
```diff
@@ -0,0 +1,138 @@
...
+  if (!file.exists(path)) {
+    stop("Failed to create local cache ('", path, "')", call. = FALSE)
+  }
+
+  "!DEBUG cache exists: `path`"
+
+  # Protect cache as much as possible
+  Sys.chmod(path, "0600")
+
+  ## TODO(jennybc): this only looks in exact directory of cache path, not up
```
jennybc: To really do this right, you need a notion of project and to ask if the cache file appears to be in a project that is an R package or git repo.

R/oauth-cache.R
```diff
@@ -0,0 +1,138 @@
...
+  df
+}
+
+entibble <- function(token) {
+  "!DEBUG entibble"
+  tibble::tibble(
+    hash = token$hash(),
+    token = list(token)
+  )
+}
```
jennybc: This is just a first pass. More variables should be pulled out.

R/oauth-cache.R
```diff
@@ -0,0 +1,190 @@
...
+  tokens <- tokens[matches, ]
+
+  if (nrow(tokens) == 1) {
+    "!DEBUG Using a token cached for `extract_email(names(tokens))`"
+    return(tokens$token[[1]])
+  }
+
+  ## TODO(jennybc) if not interactive? just use first match? now I just give up
+  if (!interactive()) {
+    message("Multiple cached tokens exist. Unclear which to use.")
```
hadley: I _think_ this should `stop()`

R/oauth-cache.R
```diff
@@ -0,0 +1,190 @@
...
+    return(NULL)
+  }
+
+  emails <- extract_email(tokens$hash)
+  cat("Multiple cached tokens exist. Pick the one you want to use.\n")
+  cat("Or enter '0' to obtain a new token.")
+  this_one <- utils::menu(emails)
+
+  if (this_one == 0) return(NULL)
```
hadley: And this too?

R/Gargle-class.R
hadley: Maybe call it `add_token_to_cache()`?

hadley: Or maybe `token_cache()`

R/oauth-cache.R
jennybc: OK will do.

R/oauth-cache.R
jennybc: This one can't `stop()` because it's the returned `NULL` that sets off the OAuth dance, i.e. it's how you can get a new token for an email, in the presence of existing tokens.

R/Gargle-class.R
jennybc: renamed to `token_cache()`

R/Gargle-class.R
```diff
@@ -22,10 +32,10 @@ gargle2.0_token <- function(email = NULL,
                             scope = NULL,
                             user_params = NULL,
                             type = NULL,
-                            use_oob = getOption("httr_oob_default"),
+                            use_oob = getOption("gargle.oob_default"),
                             ## params end
                             credentials = NULL,
-                            cache = getOption("httr_oauth_cache"), ...) {
+                            cache = if (is.null(credentials)) getOption("gargle.oauth_cache") else FALSE, ...) {
```
jennybc: It is helpful to *be able* to cache tokens that hold directly-provided credentials, though you often don't want to. This is entirely about testing.

R/Gargle-class.R
```diff
@@ -99,115 +110,33 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
...
-      normalize_scopes(self$params$scope)
-    ), NULL)
-
-    # for compatibility with digest::digest()
-    msg <- paste(openssl::md5(msg[-(1:14)]), collapse = "")
-
-    # append the email
-    paste(msg, self$email, sep = "-")
+    msg <- rhash(list(self$app, self$params$scope))
+    paste(msg, self$email, sep = "_")
```
jennybc: I no longer include endpoint here, ok? Why not? because it's always google here. Why I even ask? Because httr includes endpoint in the hash (but has no notion of email).

jennybc:
This is ready for another look @hadley @craigcitro. I realize you are probably both doing Thanksgiving now.

I might keep working on it but it feels like time to finalize and merge this. It's getting pretty huge.

The only reason the build fails is mysterious codecov reasons about a missing base.

jennybc:
Post-Thanksgiving ping 🦃

R/three-legged-oauth.R
```diff
@@ -1,8 +1,23 @@
...
+#' Get an OAuth token for a user
+#'
+#' @description Consults the token cache for a suitable OAuth token and, if
+#'   unsuccessful, gets a token via the browser flow. A cached token is suitable
+#'   if it's compatible with the user's request in this sense:
+#'   * OAuth app must be same.
+#'   * Scopes. This is a check for inclusion, not exact equality.
+#'   * Email, if provided, must be same.
+#'
+#' @description If multiple suitable tokens are found, user is presented with a
```
hadley: You only need one `@description`

R/Gargle-class.R
jennybc: The hash also includes normalized scopes and email. I think it's still a pretty handy way to look for an exact match.

R/Gargle-class.R
```diff
@@ -1,31 +1,40 @@
...
+#' @return An object of class [Gargle2.0], either new or loaded from the cache.
 #' @export
 gargle2.0_token <- function(email = NULL,
                             app = gargle_app(),
                             ## params start
                             scope = NULL,
                             user_params = NULL,
                             type = NULL,
-                            use_oob = getOption("httr_oob_default"),
+                            use_oob = getOption("gargle.oob_default"),
```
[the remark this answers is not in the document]
jennybc: This is a carbon copy of httr's behaviour and docs, with the exception of the option name. In this case, I thought following httr would be helpful for people. But I can encapsulate this in a function.

R/oauth-cache.R
```diff
@@ -0,0 +1,211 @@
...
+  if (!file.exists(path)) {
+    stop("Failed to create local cache ('", path, "')", call. = FALSE)
+  }
+
+  "!DEBUG cache exists: `path`"
+
+  ## owner can read and write, but not execute; no one else can do anything
+  Sys.chmod(path, "0600")
+
+  desc <- file.path(cache_parent, "DESCRIPTION")
```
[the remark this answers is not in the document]
jennybc: I just read usethis's DESCRIPTION. It brings quite a lot along with it.

R/oauth-cache.R
```diff
@@ -0,0 +1,211 @@
...
+#     existing token (which may, in fact, have more)
+token_multi_match <- function(existing, candidate) {
+  m_email <- if (is.null(candidate$email)) {
+    TRUE
+  } else {
+    existing$email == candidate$email
+  }
+  m_app <- existing$app_hash == rhash_app(candidate$app)
+  m_scopes <- scope_ok(existing$scopes, candidate$params$scope)
+  which(m_email & m_app & m_scopes)
```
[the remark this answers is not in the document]
jennybc: `gargle_token$hash()` includes app, normalized scopes and email.

The `app_hash` in the tibble of tokens includes just the app. I wanted human-inspectable variables for email, scope, app (plus list column of the actual token objects). But OAuth app objects aren't very human-readable and this way I get a character variable and can easily scan for tokens that, say, all use the same app.

R/oauth-cache.R
```diff
@@ -0,0 +1,211 @@
...
+  if (!file.exists(cache_path) || file.size(cache_path) == 0) {
+    list()
+  } else {
+    readRDS(cache_path)
+  }
+}
+
+add_email_scope <- function(scope = NULL) {
+  scope <- scope %||% character()
+  if (any(scope == "email")) {
```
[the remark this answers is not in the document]
jennybc: Slick.

R/three-legged-oauth.R
jennybc: But I want to keep this text in the Description vs Details:

https://github.com/jeroen/commonmark/issues/7

hadley: A roxygen tag affects all text until the next tag - so the first `@description` should be adequate.

jennybc: I stand corrected! I wonder what has changed? Because you're right -- I don't need this now but I did at the time of that issue 🤔 Removing the second `@description` tag here.

PR #32, opened 2017-11-27
Functions to build and make requests; closes #31

### LINKED ISSUE #31: Extract generate_request() and friends out of googledrive
Do what is necessary to extract the functions in these files for use with suitably prepared API specs besides Drive:

  * Request generation: <https://github.com/tidyverse/googledrive/blob/master/R/generate_request.R>
  * Make request: <https://github.com/tidyverse/googledrive/blob/master/R/make_request.R>
  * Process response: <https://github.com/tidyverse/googledrive/blob/master/R/process_response.R>

jennybc:
Closes #31 Extract generate_request() and friends out of googledrive

*I pulled this out of #30, as its own PR. It is much simpler to merge. Will rewrite the history on #30 when I return there.*


jennybc:
I will merge this soon and keep going. The review request is more of an FYI, although feedback always welcome.

Will do processing responses separately.

R/develop-request.R
```diff
@@ -0,0 +1,218 @@
...
+check_params <- function(provided, spec) {
+  required <- Filter(function(x) isTRUE(x$required), spec)
+  missing <- setdiff(names(required), names(provided))
+  if (length(missing)) {
+    stop_collapse(
+      c("Required parameter(s) are missing:", missing)
+    )
+  }
+
+  unknown <- setdiff(names(provided), names(spec))
```
craigcitro: Do you need to only do this if `additionalProperties: true` isn't in the disco doc?

R/make-request.R
```diff
@@ -0,0 +1,38 @@
+#' Make a Google API request
+#'
+#' Intended primarily for internal use in client packages that provide
+#' high-level wrappers for users. `make_request()` does very little: calls an
+#' HTTP method, only adding a user agent. Typically the input is created with
+#' [`build_request()`] and the output is processed with [`process_response()`].
+#'
+#' @param x List. Holding the components for an HTTP request, presumably created
+#'   with [`generate_request()`] or [build_request()]. Must contain the a
```
craigcitro: wildly nitpicky: missing backticks around `build_request()`

R/make-request.R
```diff
@@ -0,0 +1,38 @@
...
+#'   `method` and `url`. If present, `body` and `token` are used.
+#' @param ... Optional arguments passed through to the HTTP method.
+#'
+#' @return Object of class `response` from [httr].
+#' @export
+#' @family requests and responses
+make_request <- function(x, ..., ua = gargle_ua()) {
```
craigcitro: Minor thing: unless it's already common in `httr`, I'd say it's worth calling this `user_agent` for readability.

R/utils.R
```diff
@@ -2,7 +2,66 @@
   if (length(a) > 0) a else b
 }
 
+#' @importFrom glue glue glue_data collapse
 #' @importFrom magrittr %>%
 NULL
 
 is_string <- function(x) is.character(x) && length(x) == 1
+
+commapse <- function(...) paste0(..., collapse = ", ")
```
craigcitro: 👏

R/develop-request.R
jennybc: This is my first contact with `"additionalProperties"`, so I will educate myself about that.

R/make-request.R
jennybc: OK I sorted out a few backtick and link issues.

R/make-request.R
jennybc: done

R/develop-request.R
jennybc: http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.5.6

jennybc: Relevant if we went this far (which I propose we do not, at this time):

https://cran.r-project.org/web/packages/jsonvalidate/vignettes/jsonvalidate.html

NAMESPACE
```diff
@@ -2,6 +2,7 @@
 
 export(Gargle2.0)
 export(GceToken)
+export(build_request)
```
hadley: Would it be better to have `request_build()`, `request_develop()`, `request_make()` ?

R/develop-request.R
```diff
@@ -0,0 +1,218 @@
+#' Build a Google API request
+#'
+#' Intended primarily for internal use in client packages that provide
+#' high-level wrappers for users.
+#'
+#' @param endpoint List of information about the target endpoint. Presumably
+#' prepared from the [Discovery
```
hadley: Missing indents for a couple of params

R/develop-request.R
```diff
@@ -0,0 +1,218 @@
...
+#' @param body List. Values to send in the API request body.
+#' @param key API key. Needed for requests that don't contain a token. For more,
+#'   see Google's document [Credentials, access, security, and
+#'   identity](https://support.google.com/googleapi/answer/6158857?hl=en&ref_topic=7013279).
+#'   A key can be passed as a named component of `params`, but note that the
+#'   formal argument `key` will clobber it, if non-`NULL`.
+#' @param token Token, ready for inclusion in a request, i.e. prepared with
+#'   [httr::config()].
+#'
+#' @section develop_request():
```
hadley: Should this be in backticks?

R/develop-request.R
```diff
@@ -0,0 +1,218 @@
...
+    token = token
+  )
+
+  out$url <- httr::modify_url(
+    url = base_url,
+    path = out$path,
+    ## prevent a trailing `?` or `?=` when the query is trivial, e.g. list() or
+    ## contains a single element which is NULL
+    ## https://github.com/r-lib/httr/issues/451
+    ## could simply use out$query this if required httr >= 1.3.1
```
hadley: I think we could just require 1.3.1 or greater.

R/make-request.R
```diff
@@ -0,0 +1,39 @@
...
+#'
+#' Intended primarily for internal use in client packages that provide
+#' high-level wrappers for users. `make_request()` does very little: calls an
+#' HTTP method, only adding a user agent. Typically the input is created with
+#' [build_request()] and the output is processed with `process_response()`.
+#'
+#' @param x List. Holding the components for an HTTP request, presumably created
+#'   with [develop_request()] or [build_request()]. Must contain the a `method`
+#'   and `url`. If present, `body` and `token` are used.
+#' @param user_agent A user agent string, prepared by [httr::user_agent()].
```
hadley: I think this should include a little advice: We recommending including the name of your package and its version in the user agent.

R/make-request.R
```diff
@@ -0,0 +1,39 @@
...
+#' @param ... Optional arguments passed through to the HTTP method.
+#'
+#' @return Object of class `response` from [httr].
+#' @export
+#' @family requests and responses
+make_request <- function(x, ..., user_agent = gargle_user_agent()) {
+  method <- list(
```
hadley: A `switch()` would be slightly more elegant here and would make it easier to throw an error if the method did not match.

NAMESPACE
jennybc: It seems so obvious now that you say it. Yes.

R/develop-request.R
jennybc: Yes, it should but I wrote this right after making the usethis website:

http://usethis.r-lib.org/reference/ci.html

Yes I should just backtick it and assume all the presentation formats will get sorted out.

NAMESPACE
jennybc: done

R/develop-request.R
jennybc: fixed

R/develop-request.R
jennybc: done

R/make-request.R
jennybc: done

PR #40, opened 2018-08-22
Refactor OAuth token class and the caching strategy

jennybc:
Top priority: make email (or user or account) a first-class data field for Google OAuth tokens.

For many Google APIs, you regularly switch between identities (jane@rstudio.com vs jane@gmail.com vs ...). httr's Token2.0 class makes this extremely difficult because it has no notion of email/account and only "cares" about endpoint, app, scope:

https://github.com/r-lib/httr/blob/976289a3596dc01dc994f8fd743770a172bbecdb/R/oauth-token.r#L103-L111

``` r
hash = function() {
    # endpoint = which site
    # app = client identification
    # params = scope
    msg <- serialize(list(self$endpoint, self$app, self$params$scope), NULL)

    # for compatibility with digest::digest()
    paste(openssl::md5(msg[-(1:14)]), collapse = "")
  }
```

The `gargle2.0` class subclasses `httr::Token2.0` and makes the email address a first-class data field.

The old `$hash()` is now `$hash_short()` and `$hash()` catenates `$hash_short()` and the email address.

``` r
  hash_short = function() {
    rhash(list(self$endpoint, self$app, self$params$scope))
  },
  hash = function() {
    paste(self$hash_short(), self$email, sep = "_")
  }
```

An old hash: `2a46e6750476326f7085ebdab4ad103d`

New hashes for same oauth endpoint w/ same app and scopes but as different users:
`2a46e6750476326f7085ebdab4ad103d_jane@rstudio.com`
`2a46e6750476326f7085ebdab4ad103d_jane@gmail.com`

---

Secondary priority: it's common to hit the same Google API from multiple projects, so we want centralized, user-level management of Google tokens. They are also less likely to be pushed to the cloud if we get them out of the projects. This is a good first step towards handling them more securely, e.g. via keyring.

This PR implements a user-level cache for OAuth tokens, by default, in `~/.R/gargle/gargle-oauth`.


R/Gargle-class.R
```diff
@@ -99,115 +107,46 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
...
-      self$app,
-      normalize_scopes(self$params$scope)
-    ), NULL)
-
-    # for compatibility with digest::digest()
-    msg <- paste(openssl::md5(msg[-(1:14)]), collapse = "")
-
-    # append the email
-    paste(msg, self$email, sep = "-")
+    paste(self$hash_short(), self$email, sep = "_")
```
hadley: I think you can just use `super$hash()` here? (Then you don't need `hash_short()`)

R/oauth-cache.R
```diff
@@ -0,0 +1,245 @@
+## this file has its origins in oauth-cache.R in httr
+## nothing there is exported, so copied over, then it evolved
+
+# cache --------------------------------------------------------------
+
+gargle_default_oauth_cache_path <- "~/.R/gargle/gargle-oauth"
```
hadley: Should we be using keyring for this?

R/oauth-cache.R
```diff
@@ -0,0 +1,245 @@
...
+cache_establish <- function(cache = getOption("gargle.oauth_cache")) {
+  if (length(cache) != 1) {
+    stop("Cache should be length 1 vector", call. = FALSE)
+  }
+  if (!is.logical(cache) && !is.character(cache)) {
+    stop("Cache must either be logical or string (file path)")
+  }
+
+  # If NA, propose default cache file
+  # request permission to create it, if doesn't exist yet
```
hadley: Do we need to handle the case where the user can't write to that file?

R/oauth-cache.R
```diff
@@ -0,0 +1,245 @@
...
+
+  if (anyDuplicated(nms)) {
+    stop("Duplicate tokens!, call. = FALSE")
+  }
+
+  existing
+}
+
+## useful to jennybc during development
+cache_show <- function(path = NULL) { # nocov start
```
hadley: Maybe eventually make `cache_sitrep()` and export?

R/three-legged-oauth.R
```diff
@@ -22,6 +36,9 @@ credentials_user_oauth2 <- function(scopes,
                                     app = gargle_app(),
                                     ...) {
   "!DEBUG trying credentials_user_oauth2"
+  if (missing(scopes)) {
+    scopes <- "email"
```
hadley: Just put in args?

R/oauth-cache.R
jennybc: Yes, probably, I just couldn't bear to think about that yet.

R/oauth-cache.R
jennybc: Well, it is "handled" I think because we will soon fail to create the cache in an obvious way. User can then try again, specifying the path to a place where they **can** write and if that's a long-term preference, they can store that in the option `gargle.oauth_cache`.

Do I understand your question?

R/oauth-cache.R
jennybc: Yeah, something like that is where this heading. You will also need to be able to clear your cache or make targeted deletions.

R/Gargle-class.R
jennybc: AHA

R/three-legged-oauth.R
jennybc: I feel like I had a reason for doing this (it goes back to the ancient origins of this PR) but damn if I can remember. I'll make a comment so I re-examine when I'm really looking at general token fetching vs. just focused on oauth.

R/oauth-cache.R
hadley: Yes — I think I was just confused by "request permission", but you mean from the user.

PR #44, opened 2018-11-12
Require user to authorize token retrieval

### LINKED ISSUE #41: Confirm with user retrieving token from central store
Possibly by always asking who do you want to be.

Will need option to explicitly opt-out of the interrogation for non-interactive usage.

jennybc:
Closes #41 Confirm with user retrieving token from central store

I was able to achieve this within the existing surface. It's 75% natural / 25% awkward and I'd appreciate feedback. Because maybe some aspect of this will carry forward to httr2?

---

I explicitly used httr's cache behaviour as a reference:

  * The token-getter `gargle2.0_token()` has an argument -- `email` or `cache`, respectively -- that accepts user input but defaults to an option -- `gargle.oauth_email` or `httr_oauth_cache`, respectively.
    - `gargle.oauth_email` is not necessarily defined and therefore defaults to `NULL`. That's intentional.
    - `httr_oauth_cache` is defined in `httr:::.onLoad()` and defaults to `NA`.
  * At one extreme, `email` could provide an actual Google identity, just like providing a specific cache path.
  * At the other extreme, `email` is not specified, i.e. it's `NULL` and, one way or another, the user will be forced to authorize the use of a specific Google identity. That could happen in R or in the account choose presented by Google.
    - Difference with httr's cache: default value of `httr_oauth_cache` is `NA`, but I need `NA` for another purpose.
    - `email = NA` will match no existing email/token and force the browser dance.
  * In the middle, `email = TRUE` does not pass a specific identity but it authorizes gargle to auto-discover one. This is the awkwardness. `cache = TRUE` is a more natural expression of "yes, cache, but you can sort out the details" than the analogous `email = TRUE`.


jennybc:
Some usage

``` r
> token <- gargle2.0_token(
+   app = tidyverse_app(),
+   scope = "https://www.googleapis.com/auth/drive"
+ )
Suitable cached token(s) available.
Pick the account to use or enter '0' to obtain a new token.
1: jenny.f.bryan@gmail.com

Selection: 1
> token <- gargle2.0_token(
+   app = tidyverse_app(),
+   scope = "https://www.googleapis.com/auth/drive",
+   email = TRUE
+ )
Using cached token for jenny.f.bryan@gmail.com
> token <- gargle2.0_token(
+   scope = "https://www.googleapis.com/auth/drive",
+   email = "toby@example.com"
+ )
Press Esc/Ctrl + C to abort
Authentication complete.
```

jennybc:
"Suitable" is a euphemism for token match on OAuth endpoint, OAuth app, and scopes. I think the user we are trying to inform/protect here doesn't even know what these are. I think they are thinking "Hey, I want to sign in as jane@example.com and access my stuff on Drive". I'm skeptical that a more detailed message helps them.

I guess I *could* include the humane form of scopes that is revealed by the print method. For inspiration, here's how `Gargle2.0` tokens print:

```
> cache_show()
Reading from cache in '~/.R/gargle/gargle-oauth'
$`42819459f2dd6a26ddd02157588a2451_stella@example.org`
<Token (via gargle)>
  <oauth_endpoint> google
       <oauth_app> gargle-demo
           <email> stella@example.org
          <scopes> email, mouthwash
     <credentials> 
---
$`48e7e765f60fbe16a9bfca871e3c62a0_blah-blah@gmail.com`
<Token (via gargle)>
  <oauth_endpoint> google
       <oauth_app> gargle-demo
           <email> blah-blah@gmail.com
          <scopes> email, ...drive
     <credentials> access_token, expires_in, refresh_token, scope, token_type, id_token
---

$`d9443dbd0673bf6cb64f8258f8060d52_blah-blah@gmail.com`
<Token (via gargle)>
  <oauth_endpoint> google
       <oauth_app> tidyverse
           <email> blah-blah@gmail.com
          <scopes> email, ...drive
     <credentials> access_token, expires_in, refresh_token, scope, token_type, id_token
---
```

You're right:mI'm not caching anything about this email in the session.

The client package, once it gets this token, stewards it. We never hit this code again unless the user explicitly requests a new token with a different combination of OAuth app, scopes, and/or email. In which case, it seems appropriate that they revisit the same logic and consent choices.

jennybc:
I should clarify that `Gargle2.0` tokens also inherit httr-like caching behaviour, just with a new default to a user-level cache. I used the `cache`-handling code as inspiration for `email` handling. It's not an "either or" situation.

hadley:
In that case, let's just beef up the message a bit:

* Before: Suitable cached token(s) available.
* After:  The foo package is requesting access to your google account.

This would require a new package param specifically for use in the message, but I think that's ok?

hadley:
No, just an argument.

PR #47, opened 2018-11-21
MVP for AuthState

### LINKED ISSUE #43: Figure out how much of the auth state management can move over here
Each client package needs to track its own state but the functions for managing state could be in gargle. At least, I think so.

jennybc:
WIP to close #43 Figure out how much of the auth state management can move over here

This is the opening move in the larger plan we discussed Monday:
I create an R6 class `AuthState` here in gargle to store the auth state of a client package.

I have companion PRs: https://github.com/tidyverse/googlesheets4/pull/15 and https://github.com/tidyverse/googledrive/pull/236.

In these PRs, I remove the previous solution, where each package maintained an auth state via a private environment.

There's still work to be done (e.g. actually storing the state in gargle vs. googledrive and googlesheets4), but it's a first step!

I will keep going but feedback is welcome.


jennybc:
Re: package name. Yeah that sort of thing will happen in the next pass. I’m going to hold on to more info here and store in gargle, for a cross-package back channel.

Issue #36, opened 2018-03-29
Response processing + error message helper


hadley:
Seems like we should encode all our knowledge about error messages in one place. bigquery currently has this:

```R
signal_reason <- function(reason, message) {
  if (is.null(reason)) {
    stop(message, call. = FALSE)
  } else {
    cl <- c(paste0("bigrquery_", reason), "error", "condition")
    message <- paste0(message, " [", reason, "]")

    cond <- structure(list(message = message), class = cl)
    stop(cond)
  }
}
```

craigcitro:
One more relevant link from the sidelines: there's a published description of [how new APIs do errors](https://cloud.google.com/apis/design/errors), which includes the canonical [error codes and http mappings](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto). This view of errors is slightly newer than BQ, but it's what BQ and other APIs will ultimately converge on.

jennybc:
I am happy to work on this. Re googlesheets4, I've just gotten to the "request processing" stage (vs. request prep, request execution).

I will try to make a PR in the same spirit as https://github.com/r-lib/gh/pull/72/files that exposes everything we know, in a way that suggest how to resolve the problem.

jennybc:
@craigcitro So this source file is the definitive source of Google API error codes? 🤔

https://github.com/googleapis/googleapis/blob/ba1108b3c40c1a674158217f680dac61dd3930f6/google/rpc/code.proto#L26

jennybc:
> This view of errors is slightly newer than BQ, but it's what BQ and other APIs will ultimately converge on.

I'm starting to understand what this means. I suspect that, today, the JSON that comes back for errors has different structure for each API. I'm not sure how much can be shared, other than the tibble below, derived from the [Cloud APIs Errors doc](https://cloud.google.com/apis/design/errors). Sheets v4 appears to send this RPC code and a short description, i.e. we don't gain anything from this table.

But it's possible that one of the other APIs does not and lookup in central object in gargle would be helpful.

``` r
## https://cloud.google.com/apis/design/errors
## https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
oops <- tibble::tribble(
  ~HTTP,                  ~RPC, ~Description,
    200,                  "OK", "No error.",
    400,    "INVALID_ARGUMENT", "Client specified an invalid argument. Check error message and error details for more information.",
[...]

PR #65, opened 2019-04-19
Revise/add functions to get userinfo and tokeninfo

### LINKED ISSUE #22: Uniform way to see the email attached to any token
E.g., OAuth and service

### LINKED ISSUE #26: Establish if email scope can be used with a service token

jennybc:
Closes #22, closes #26


jennybc:
@craigcitro This is a general upgrade of my brain and the function(s) for introspecting on a token. The only immediate and direct use is that we use the email heavily now when indexing cached OAuth tokens.

You can see I've also used the comments as a place to park some notes.

I'm asking for your review just in terms of general perspective and folklore. Do you think the `get_email()` function should rely on userinfo or tokeninfo?

R/get-email.R
```diff
@@ -1,18 +1,76 @@
...
+    method = "GET",
+    path = "v1/userinfo",
+    token = token,
+    base_url = "https://openidconnect.googleapis.com"
+  )
+  raw_resp <- gargle::request_make(req)
+  ## TODO(jennybc) use process_response() once it's available in same pkg
+  httr::stop_for_status(raw_resp)
+  res <- httr::content(raw_resp, as = "parsed", type = "application/json")
+  #res$email
```
craigcitro: silly: drop

R/get-email.R
```diff
@@ -1,18 +1,76 @@
...
+  res
+}
 
-  res <- httr::GET(url)
+# WARNING: does not validate or refresh token!
```
craigcitro: I guess I'm a little confused -- this does *validate* it (with the server), right?

R/get-email.R
```diff
@@ -1,18 +1,76 @@
...
+# Here's the OpenID Connect discovery document:
+# https://accounts.google.com/.well-known/openid-configuration
+#
+# Here's the URL for userinfo_endpoint, at the time of writing:
+# https://openidconnect.googleapis.com/v1/userinfo
+#
+# I had working code to access both of these endpoints before I read this and
+# there are a few differences between what I do and what they say to do. The
+# discrepancies are around which token to send (ID? access? etc.) and where/how
+# to send it. I conclude that there must be a few variations that work.
```
craigcitro: So id tokens are a different way to prove your identity, as they're self-signed tokens that anyone can verify (and implement the openid connect "spec").

Some details are [in the docs](https://developers.google.com/identity/protocols/OpenIDConnect#exchangecode), but roughly speaking, this is my understanding:

* if I grant the `userinfo.email` scope, the service I'm talking to can directly talk to google to determine my identity

* if I create an id token, I can hand that to a service, and they can verify with google that this is a valid token

Both play similar roles; my impression is that the first is more useful for us here, since we're in the business of talking to one or more google APIs, and the second is more useful in a situation where Google is just one of many identity providers a server may want to use.

I don't have a huge amount of experience with id tokens, but one use case (that may be interesting for Shiny?) that's worth calling out: if you have both a frontend and a backend, and the user is going to authenticate to your frontend, it's nice to avoid them doing another auth step against your backend. One way to make this work is to have your frontend get an id token, which you then present to the backend with the request; the backend can verify the identity based on that. (However, this only solves a problem if the backend doesn't need additional scopes, or is willing to walk you through additional flows for them later.)

R/get-email.R
jennybc: Yeah, that will not survive to the point where I merge this. I'm still playing around with these functions.

R/get-email.R
jennybc: True. But it does not validate proactively or in a way that produces a actionable error message if your token has grown stale. You just get a 400 Bad Request* and you have to remember that the explanation might just be that you need to refresh. I'll reword to make this more clear to Future Me.

*or maybe once I do #36, I'll see there's a beautifully informative message coming back

R/get-email.R
jennybc: Thanks for this. Glad you also agree that userinfo.email is the more useful route for us. I'll file away this knowledge about ID tokens and possible relevance to Shiny.

R/get-email.R
jennybc: This conversation made a 💡go on for me: I think the "email fetching" function should route through `get_userinfo()`, not `get_tokeninfo()`, because the former naturally exploits the built-in refresh feature of our tokens.

jennybc: Or, at least, this is an important difference to note.

PR #67, opened 2019-04-22
Introduce gargle_quiet option + a central output printer

### LINKED ISSUE #66: Debug mode

jennybc:
Closes #66

This is a _very low tech_ solution for getting a "debug" or "chatty" mode. It just uses one function, `cat_line <- function(..., quiet = gargle_quiet())`, where `gargle_quiet()` does `getOption("gargle_quiet", default = TRUE)`.

I remove a more proper solution, using debugme, but I never managed to achieve a smooth dev workflow with it. Currently you need to: set an env var, unload debugme, re-load debugme, reload the target package every time you toggle debug mode. It should be fairly easy to return debugme in the future, if it becomes easier to use.

I am unsure whether I should really be doing this via `cat()` (vs. `message()`) but that is also easy enough to change. I just want to set this in motion and start enjoying greater visibility during development.

A general reorg of UI-oriented functions came along for the ride.


jennybc:
I'm going to merge in order to keep moving briskly, but will happily act on any feedback this PR provokes.

R/glue-conditions.R
```diff
@@ -0,0 +1,53 @@
+stop_glue <- function(..., .sep = "", .envir = parent.frame(),
```
hadley: We can talk about this in depth at some point, but for now, I don't think this approach is something we should be copying going forward.

R/ui.R
```diff
@@ -0,0 +1,50 @@
...
+  cat(glue_data(..., .sep = .sep, .envir = .envir), sep = "\n")
+}
+
+commapse <- function(...) paste0(..., collapse = ", ")
+bt <- function(x) encodeString(x, quote = "`")
+sq <- function(x) encodeString(x, quote = "'")
+
+## obscure the middle bit of (sensitive?) strings with '...'
+## obfuscate("sensitive", first = 3, last = 2) = "sen...ve"
+obfuscate <- function(x, first = 6, last = 4) {
```
hadley: Oh I like this idea!

PR #69, opened 2019-04-23
Secret management

### LINKED ISSUE #3: Provide tools and instructions for embedding an encrypted token
And include some docs about how to create, encrypt, and decrypt.

jennybc:
Closes #3


.travis.yml
```diff
@@ -17,5 +25,8 @@ matrix:
         script: Rscript -e 'pkgdown::deploy_site_github(verbose = TRUE)'
         skip_cleanup: true
     - r: oldrel
+      env: GARGLE_PASSWORD=''
     - r: 3.3
+      env: GARGLE_PASSWORD=''
     - r: 3.2
+      env: GARGLE_PASSWORD=''
```
jennybc: In bigrquery, you run auth-requiring tests on 3.2 and release. Is that because you sort of want to "bookend" things? You skip auth-requiring tests on the versions in between.

R/secret.R
```diff
@@ -0,0 +1,134 @@
+# Originally from r-dbi/bigrquery/secret.R
```
jennybc: I suppose I will turn this into an article.

R/secret.R
```diff
@@ -0,0 +1,134 @@
...
+secret_can_decrypt <- function(package) {
+  requireNamespace("sodium", quietly = TRUE) && secret_pw_exists(package)
+}
+
+secret_write <- function(package, name, data) {
+  #if (inherits(data, "connection")) {
+    data <- readBin(data, "raw", file.size(data))
+  #} else if (is.character(data)) {
+  #  data <- charToRaw(data)
+  #}
```
jennybc: The code in bigrquery doesn't have any of this commented out. But I don't see how that could actually work?

If `data` is a connection, you can't call `file.size()` on it. If `data` is a file path, we don't fulfill the condition `inherits(data, "connection")`, so we never get to `readBin()`.

My current token was encrypted with the code "as is" above, i.e. with those lines commented out, and with this call:

``` r
secret_write(
  package = "gargle",
  name = "gargle-testing.json",
  data = "~/.R/gargle/gargle-testing.json"
)
```

Do you agree the bigrquery code can't work as written unless the entire secret is passed as a string via `data`? Or what am I missing about how to call this function?

R/secret.R
```diff
@@ -0,0 +1,134 @@
...
+
+  secret <- fs::path("inst", "secret")
```
jennybc: I "fs-ified" everything, because it is also used here.

R/secret.R
```diff
@@ -0,0 +1,134 @@
...
+  attr(enc, "nonce") <- NULL
+  writeBin(enc, dst)
+
+  invisible(dst)
+}
+
+# Generated with sodium::bin2hex(sodium::random(24)). AFAICT nonces are
+# primarily used to prevent replay attacks, which shouldn't be a concern here
+secret_nonce <- function() {
+  sodium::hex2bin("cb36bab652dec6ae9b1827c684a7b6d21d2ea31cd9f766ac")
```
jennybc: Generated a new nonce for gargle.

tests/testthat/test-aaa.R
```diff
@@ -0,0 +1,14 @@
...
+
+  expect_error_free(
+    token <- credentials_service_account(
+      scopes = "https://www.googleapis.com/auth/userinfo.email",
+      path = rawToChar(secret_read("gargle", "gargle-testing.json"))
+    )
+  )
+  email <- get_email(token)
+  expect_match(email, "^gargle-testing@.*[.]iam[.]gserviceaccount[.]com")
+})
```
jennybc: The usage of the encrypted token(s) in testing is going to look pretty different in gargle vs. bigrquery (or googledrive or googlesheets4). In the wrappers, we decrypt the token once and use it in all tests that require auth. In gargle, I have a different goal, i.e. I need to verify I can load / refresh/ introspect on multiple tokens (eventually).

.travis.yml
hadley: I'm not sure what I was thinking — I suspect I meant to also not run auth on 3.2. Some more cogent thoughts:

Tests that require auth tend to be slower (since they're going to be sending data back and forth), and the API is independent of the R version, so all else being equal it seems that you only need to run on R-release. 

I think I also deliberately run on R-release without auth because that's the closest to what CRAN will see, and it seems nice to double check.

R/secret.R
hadley: Yeah, I think this must be a path that I tried earlier on, but eventually gave up on, so it become vestigial and buggy.

R/secret.R
hadley: I don't _think_ this is necessary but I'm not sure.

tests/testthat/test-aaa.R
hadley: Makes sense.

jennybc:
I'm going to merge this soon, because it is already clearly better than ... nothing!

But @Hadley note that I've developed an article around this now, in case that elicits any more feedback from you.

@jeroen and @richfitz I also welcome your input, since you've got loads of relevant experience from working on [sodium](https://github.com/jeroen/sodium#readme) and [cypher](https://ropensci.github.io/cyphr/) (which I just ?re-?discovered tonight). I care about:

1. having a reasonable approach to testing token management in gargle (and googledrive, googlesheets4, etc.)
1. helping authors of other API wrappers develop a plan for their tokens

R/secret.R
```diff
@@ -0,0 +1,89 @@
+# Setup support for the NAME=PASSWORD envvar ----------------------------------
+
+# secret_pw_name("gargle") --> "GARGLE_PASSWORD"
+secret_pw_name <- function(package) {
+  paste0(toupper(package), "_PASSWORD")
+}
+
+# secret_pw_gen() --> "9AkKLa50wf1zHNCnHiQWeFLDoch9MYJHmPNnIVYZgSUt0Emwgi"
+secret_pw_gen <- function() {
+  x <- sample(c(letters, LETTERS, 0:9), 50, replace = TRUE)
```
richfitz: Given you already have sodium pulled in as a dependency, I would be inclined to use `sodium::random` here to avoid the possibility of generating identical passwords as a result of `set.seed`.  It would be a bit of a faff though as you'd need to rejection sample its output, or do something else to map the raw bytes to your pool of characters.

vignettes/articles/managing-tokens-securely.Rmd
```diff
@@ -0,0 +1,198 @@
...
+
+### Define env var in `.Renviron`
+
+Combine the name and value to form a line like this in your user-level `.Renviron` file:
+
+```{r, echo = FALSE, results = "asis"}
+cat(paste0(pw_name, "=", pw), sep = "\n")
+```
+
+`usethis::edit_r_environ()` can help create or open this file. We recommend using the user-level `.Renviron`, as opposed to project-level, for the usual reason  that this makes it less likely you will share sensitive information by mistake. Make sure `.Renviron` ends in a newline; the lack of this is a notorious cause of silent failure. Remember you'll need to restart R for the newly defined environment variable to take effect.
```
richfitz: You quite possibly have good reasons for not mentioning it but you can also reload .Renviron with `readRenviron("~/.Renviron")`

vignettes/articles/managing-tokens-securely.Rmd
```diff
@@ -0,0 +1,198 @@
...
+
+We go ahead and attach gargle in the normal way, which makes the exported functions readily available.
+
+```{r setup}
+library(gargle)
+```
+
+## Overview of the approach
+
+  * Generate a random PASSWORD and give it a self-documenting name, e.g. `GARGLE_PASSWORD`.
```
richfitz: Here you make it sound like a naming suggestion, but it looks like your functions require the format `<package>_PASSWORD`

vignettes/articles/managing-tokens-securely.Rmd
```diff
@@ -0,0 +1,198 @@
...
+On Travis-CI, define this env var in your repo settings via the browser UI:
+
+https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings
+
+Alternatively, you can use the Travis command line interface to configure the env var or even define an encrypted env var in `.travis.yml`. In any case, remember that private env vars are not available to pull requests, which is another reason to carry on gracefully when token decryption is not possible (see below).
+
+On R-hub, send this env var in your calls to `check()` and friends:
+
+```
+rhub::check(env_vars = Sys.getenv(secret_pw_name("gargle"), names = TRUE))
```
richfitz: should this be `gargle:::secret_pw_name`?

vignettes/articles/managing-tokens-securely.Rmd
```diff
@@ -0,0 +1,198 @@
...
+  * `data`, either a file path to the unencrypted secret file or the data to
+    be encrypted as a raw vector. In the case of a secret file, it should
+    normally be stored outside your package, to decrease the chance of
+    accidental leakage.
+
+Example of a call to `secret_write()`, where `gargle-testing.json` is a JSON
+file downloaded for a service account managed via the [Google API / Cloud Platform console](https://console.cloud.google.com/project).
+
+```{r eval = FALSE}
+secret_write(
```
richfitz: `gargle:::secret_write` ?

vignettes/articles/managing-tokens-securely.Rmd
```diff
@@ -0,0 +1,198 @@
...
+
+  * Decryption is not going to work. Either because the Suggested [sodium](https://cran.r-project.org/web/packages/sodium/index.html) package is not available or (much more likely) because the env var that represents the key is not available.
+    - This will be the case on CRAN, by definition, because there is no way to share an encrypted secret.
+    - This will be the case for external contributors, on their personal machines and when their GitHub pull requests are checked via CI services, such as Travis-CI or AppVeyor. 
+  * Decryption is going to work. This is where you actually get to test package functionality against the target API.
+
+In a wrapper package, you could determine decrypt-ability at the start of the test run. Here's representative code from bigrquery's `tests/testthat/setup-auth.R` file:
+
+```{r eval = FALSE}
+if (secret_can_decrypt("bigrquery")) {
```
richfitz: `gargle:::secret_can_decrypt` and `gargle:::secret_read`

vignettes/articles/managing-tokens-securely.Rmd
jennybc: 😲 I didn't know about that function!

vignettes/articles/managing-tokens-securely.Rmd
jennybc: Yeah this one should. Thanks.

vignettes/articles/managing-tokens-securely.Rmd
jennybc: In this case, no, because it really does refer to versions of `secret_can_decrypt()` and `secret_read()` that are available in bigrquery. Which is how other wrappers would probably have to do it too because I bet `:::` would cause trouble at CRAN. I've added explanatory words.

vignettes/articles/managing-tokens-securely.Rmd
jennybc: True. Reworded.

PR #77, opened 2019-05-06
Add response_process()

### LINKED ISSUE #36: Response processing + error message helper

jennybc:
Closes #36

This is working successfully in both googledrive (https://github.com/tidyverse/googledrive/pull/256) and googlesheets4 (https://github.com/tidyverse/googlesheets4/pull/25).


NAMESPACE
```diff
@@ -28,6 +28,7 @@ export(oauth_app_from_json)
 export(request_build)
 export(request_develop)
 export(request_make)
+export(response_process)
```
jennybc: I don't love this name but we've already got a very strong `NOUN_VERB()` pattern going here and it feels very right for `credentials_*()`, `request_*()`, and `token_*()`. So it feels like I'm committed now and must see this through.

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+#' examination.
+#'
+#' @param resp Object of class `response` from [httr].
+#' @param raw Logical. Whether to return parsed content (`FALSE`) or in `raw`
+#'   form.
+#'
+#' @return The content of the request.
+#' @family requests and responses
+#' @export
+response_process <- function(resp, raw = FALSE) {
```
jennybc: Most inspiring ancestor: similar code in bigrquery.

https://github.com/r-dbi/bigrquery/blob/feb1a9ab3019bdd96848acce81cdcfe6770696d6/R/bq-request.R#L164-L206

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+  } else {
+    jsonlite::fromJSON(rawToChar(content), simplifyVector = FALSE)
+  }
+}
+
+check_response <- function(resp) {
+  code <- httr::status_code(resp)
+  # code could be in 100s, 200s, 300s, 400s, 500s
+
+  if (code < 200 || (code >= 300 && code < 400)) {
```
jennybc: I know this looks a bit odd. But I've never seen anything in the 100s or 300s from any of the Google APIs I've worked with. And I don't think it should happen?

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+  if (code >= 200 && code < 300) {
+    check_for_json(resp)
+    return()
+  }
+  # code could be: 400-499, 500-599
+
+  google_error(resp)
+}
+
+stop_request_failed <- function(resp, ..., message = NULL, .subclass = NULL) {
```
jennybc: Everything routes through here.

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+    glue("  * Non-JSON content type: {type}")
+  )
+  stop_request_failed(
+    resp,
+    message = message,
+    text = httr::content(resp, as = "text")
+  )
+}
+
+google_error <- function(resp) {
```
jennybc: What should this be called?

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+  check_for_json(resp)
+
+  code    <- httr::status_code(resp)
+  type    <- httr::http_type(resp)
+  content <- httr::content(resp, as = "raw")
+  content <- jsonlite::fromJSON(rawToChar(content), simplifyVector = FALSE)
+
+  error <- content[["error"]]
```
jennybc: This gets very if-else-y. I have identified at least 3 structurally distinct error payload styles and I'm betting there are more 😬

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+      message <- c(
+        httr::http_status(resp)$message,
+        glue("  * {format(names(errors), justify = 'right')}: {errors}")
+      )
+    }
+  }
+  stop_request_failed(resp, message = message)
+}
+
+redact_response <- function(resp) {
```
jennybc: I think the pros of this outweighs the cons. If people store a response as `.rds` (such as the test fixtures I'm adding), it seems best to redact the auth tokens.

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+#' All that's needed for a successful request is to parse the JSON extracted via
+#' `httr::content()`. Therefore, the main point of `response_process()` is to
+#' handle less happy outcomes:
+#'   * Status code in the 100s (information) or 300s (redirection). These are
+#'     unexpected.
+#'   * Non-JSON content type, such as HTML.
+#'   * Status codes in the 400s (client error) and 500s (server error). The
+#'     structure of the error payload varies across Google APIs and we try to
+#'     create a useful message for all variants we know about.
+
```
jennybc: Note to self: remove this empty line

R/response_process.R
hadley: I think we should talk about this in person, because this feels like it falls in an uncanny valley.

R/response_process.R
hadley: Actually, I think `check_for_json()` should live in `response_process()` and this should be called `check_for_success()` or similar.

R/response_process.R
hadley: I think with some other refactoring, suggested above, this should be inlined into `check_for_success()`

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+google_error <- function(resp) {
+  check_for_json(resp)
+
+  code    <- httr::status_code(resp)
+  type    <- httr::http_type(resp)
+  content <- httr::content(resp, as = "raw")
+  content <- jsonlite::fromJSON(rawToChar(content), simplifyVector = FALSE)
+
+  error <- content[["error"]]
+  if(is.null(error)) {
```
hadley:
```suggestion
  if (is.null(error)) {
```

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+  )
+  stop_request_failed(
+    resp,
+    message = message,
+    text = httr::content(resp, as = "text")
+  )
+}
+
+google_error <- function(resp) {
+  check_for_json(resp)
```
hadley: Maybe `check_for_json()` should return a value? So you could do `json <- check_for_json(resp)`?

tests/testthat/fixtures/README.Rmd
```diff
@@ -0,0 +1,70 @@
+---
```
hadley: Where does this `/fixtures/` convention come from?

jennybc: When there are lots of loose "testing parts" needed, I like to keep them in a folder like this. I do this elsewhere. And I thought I heard the word "fixtures" used for such things and liked it better than "test-files".

R/response_process.R
```diff
@@ -0,0 +1,190 @@
...
+
+#'
+#' @details
+#' A redacted version of the `resp` input is returned in the condition (auth
+#' tokens are removed). Use functions such as `rlang::last_error()` or
+#' `rlang::catch_cnd()` to capture the condition and do a more detailed forensic
+#' examination.
+#'
+#' @param resp Object of class `response` from [httr].
+#' @param raw Logical. Whether to return parsed content (`FALSE`) or in `raw`
```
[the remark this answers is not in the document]
jennybc: I actually took this feature directly from bigrquery and have never had it in the analogous functions in googledrive or googlesheets(4).

I thought about it and decided perhaps it would be useful for knocking out intermediate steps, sometimes?

A little research to inform our discussion:

I believe this is used (exactly?) once in bigrquery:

https://github.com/r-dbi/bigrquery/blob/b5b7ccf30a939ed57d722b807818197c11bb8faa/R/bq-download.R#L232

`bq_get()` is the only such request-wrapper that even exposes `raw`, so I suspect this is why the argument exists?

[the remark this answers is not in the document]
jennybc: If we remove the `raw` argument, it would be nice to refactor `response_process()` in such a way that the above bigrquery usage has another reasonable solution.

R/response_process.R
```diff
@@ -0,0 +1,165 @@
...
+#'   * Response processed with [response_process()].
+#'
+#' All that's needed for a successful request is to parse the JSON extracted via
+#' `httr::content()`. Therefore, the main point of `response_process()` is to
+#' handle less happy outcomes:
+#'   * Status codes in the 400s (client error) and 500s (server error). The
+#'     structure of the error payload varies across Google APIs and we try to
+#'     create a useful message for all variants we know about.
+#'   * Non-JSON content type, such as HTML.
+#'   * Status code in the 100s (information) or 300s (redirection). These are
```
jennybc: @craigcitro Is my impression correct? That we should not be expecting statii in the 100s or 300s. I've never seen it, but I'd love your input.

R/response_process.R
```diff
@@ -0,0 +1,165 @@
...
+
+  if (code >= 200 && code < 300) {
+    if (code == 204) {
+      # HTTP status: No content
+      TRUE
+    } else {
+      response_as_json(resp)
+    }
+  } else {
+    stop_request_failed(error_message(resp), resp)
```
jennybc: The one thing we lost in the refactor is that, when we error on a status in the 100s or 300s, we don't highlight that particular cause. And I imagine the message will be very minimal, as the content probably isn't really one of the errors I've planned around. But I also think this situation is OK until someone tells us it's not.

R/response_process.R
craigcitro: I've never seen an API return a value in the `1xx` range. I don't think that's written in stone anywhere, but I'd be really surprised.

I think `3xx` errors are possible, eg when fetching data -- BQ may be a case where you encounter this. GCS documents [this possibility](https://cloud.google.com/storage/docs/json_api/v1/status-codes).

There's also [this page](https://cloud.google.com/apis/design/errors#http_mapping) with a semi-authoritative answer about errors and status codes, though it's mostly for "newer" APIs.

jennybc: OK thanks. I have read that page and have built in some of this RPC information. I think we are OK for a first release and if our logic proves to be too simplistic, people will tell us and we will adjust in the face of some concrete examples.

PR #80, opened 2019-06-05
Secure built in assets

jennybc:
Just dealing with API keys here. OAuth apps will come in another PR.


R/inside-the-house.R
```diff
@@ -0,0 +1,22 @@
+from_permitted_package <- function(env = parent.frame()) {
+  env <- topenv(env, globalenv())
+  if (!isNamespace(env)) {
+    return(FALSE)
+  }
+
+  nm <- getNamespaceName(env)
+  cat_line("attempt from: ", nm)
```
hadley: Is this just for debugging?

R/gargle-api-key.R
```diff
@@ -33,13 +33,6 @@
...
-    "AIzaSyDXic",
-    # pgnTLHYCI
-    "V6oK4vR948",
-    # HCWveSQkJ
-    "PUAvlyr7Qh",
-    # ddS4QrEge
-    "kNwCoQ0cA"
-  )
+  check_permitted_package(parent.frame())
+  gak
```
hadley: Can you turn `gak` and `tak` into functions? That will save one step when we introduce greater obfuscation

R/inside-the-house.R
```diff
@@ -0,0 +1,22 @@
...
+  nm %in% c("gargle", "googledrive", "googlesheets4", "gmailr", "bigrquery")
+}
+
+check_permitted_package <- function(env = parent.frame()) {
+  if (!from_permitted_package(env)) {
+    msg <- paste(
+      "Resource is only available inside specific tidyverse packages.",
```
hadley: Would it be better to frame this as "gargle project"? Since that's what you'll need to create in the cloud console?

R/inside-the-house.R
jennybc: Yes.

R/inside-the-house.R
jennybc: I don't quite follow. The user (or package author) who runs into this doesn't need to create a "gargle project" *per se*. They just need to bring their own API key or OAuth client id/secret pair. Is this just about proposing a project name for them?

jennybc: (And yes those things must live inside a GCP project). But the connection between the user's GCP project (and its assets around credentials) and gargle would only be a state of mind.

R/gargle-api-key.R
jennybc: ok done

R/inside-the-house.R
hadley: I think I mean "google project" 😞

tests/testthat/test-inside-the-house.R
```diff
@@ -0,0 +1,12 @@
+test_that("gargle is 'inside the house'", {
+  expect_true(from_permitted_package())
+  expect_error_free(check_permitted_package())
+})
+
+test_that("it is possible to be 'outside the house'", {
+  expect_false(local(gargle:::from_permitted_package(), envir = globalenv()))
```
hadley: I don't _think_ you need `gargle:::` here

R/gargle-api-key.R
```diff
@@ -4,16 +4,15 @@
 #' which case the request must be sent with an API key in lieu of a token. This
 #' function provides an API key for limited use in prototyping and for testing
 #' and documentation of gargle itself. This key may be deleted or rotated at any
-#' time. This function is not exported and will only work inside other gargle
-#' functions, such as [request_build()].
+#' time. There are no guarantees about which APIs are enabled.
```
hadley: *DO NOT USE IN A PACKAGE*

PR #81, opened 2019-06-05
Convenience method for creating srcrefs

hadley:
To be used as follows:

```R
fun <- function() "Hello"
attr(fun, "srcref") <- new_srcref(
  "The source code of this function is concealed"
)

fun
#> The source code of this function is concealed
```

I don't have a great sense of what text we should use; we should probably emphasise that the contents of the function are for internal use only.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -64,3 +64,12 @@ add_email_scope <- function(scopes = NULL) {
   url <- "https://www.googleapis.com/auth/userinfo.email"
   union(scopes %||% character(), url)
 }
+
+new_srcref <- function(lines) {
+  n <- length(lines)
+
+  srcref(
+    srcfilecopy("HIDDEN", lines),
+    c(1L, 1L, n, nchar(lines[[n]]))
+  )
+}


jennybc:
Can't someone always just look at the `body()`?

``` r
the_real_thing <- function() invisible("SUPER_SECRET") 

the_real_thing
#> function() invisible("SUPER_SECRET")

the_real_thing()

body(the_real_thing)
#> invisible("SUPER_SECRET")

scold <- function() "Avert your eyes!"
attr(the_real_thing, "srcref") <- attr(scold, "srcref")

the_real_thing
#> function() "Avert your eyes!"

the_real_thing()

body(the_real_thing)
#> invisible("SUPER_SECRET")
```

<sup>Created on 2019-06-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

hadley:
If you’re concerned about that, we could use `{}` and also obscure the srcref of that. But there’s no trick that can stop everything — you could also just `debug()` the function. I think the point is just to avoid anyone accidentally revealing the app info.

PR #98, opened 2019-07-24
Gce fixes

### LINKED ISSUE #97: Can't get existing credentials_gce() to work
----
We are trying to use the `biqrquery` package on a Google VM, where we obtain the needed token from the VM metadata service, i.e. using just the function `credentials_gce()`. In our use case, minting service account keys is not an option. Despite trying various approaches, we were unable to get it to work, and would get the error messages:

```
Error: Can't get Google credentials.
Are you running bigrquery in a non-interactive session? Consider:
  * Call `bq_auth()` directly with all necessary specifics.
Execution halted
```

Here is an example that creates the above  messages:
```r
library(gargle)
library(bigrquery)
projectID = "my-cloud-project"
cred_funs_clear()
cred_funs_add(credentials_gce)
scopes = c("https://www.googleapis.com/auth/bigquery", "https://www.googleapis.com/auth/cloud-platform")
bq_auth(scopes)
```

We tracked the issue down into `gargle` and tweaked `credentials_gce()` a bit to get it working. We can submit a pull request if this might be useful? Thx.

wlongabaugh:
Closes #97 

Fixes to credentials_gce() and related functions so it does not return NULL.

### DIFF
--- a/R/credentials_gce.R
+++ b/R/credentials_gce.R
@@ -74,6 +74,7 @@ GceToken <- R6::R6Class("GceToken", inherit = httr::Token2.0, list(
     # The access_token can only include the token itself, not the expiration and type. Otherwise, the
     # httr code will create extra header lines that bust the POST request:
     full_token <- fetch_access_token(self$params$scope, service_account=self$params$service_account)
+    self$credentials <- list(access_token = NULL)
     self$credentials$access_token <- full_token$access_token
   },
   revoke = function() {}
@@ -133,7 +134,8 @@ get_instance_scopes <- function(service_account) {
   # Gotta have the "/" in front of the scopes to make the URL well-formed:
   path <- paste0("instance/service-accounts/", service_account, "/scopes")
   scopes <- gce_metadata_request(path)
-  ct <- httr::content(scopes, as = "text")
+  # Add the encoding parameter to quiet warning message about UTF8:
+  ct <- httr::content(scopes, as = "text", encoding = "utf8")
   strsplit(ct, split = "\n", fixed = TRUE)[[1]]
 }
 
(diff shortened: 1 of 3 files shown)


R/credentials_gce.R
```diff
@@ -27,13 +27,22 @@ credentials_gce <- function(scopes = NULL, service_account = "default", ...) {
   if (!all(scopes %in% instance_scopes)) {
     return(NULL)
   }
-  credentials <- fetch_access_token(scopes, service_account = service_account)
+
+  # The access_token can only include the token itself, *not the expiration and type*. Otherwise, the
+  # httr code will create extra header lines that bust the POST request:
```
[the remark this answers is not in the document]
wlongabaugh: Sure. If I do not trim down the credentials to just the token, I head up with a POST header that looks like:
```
Authorization1: Bearer (hex token characters)
Authorization2: Bearer (expiration seconds)
Authorization3: Bearer Bearer
```
So the credentials can only have the token to get a correctly formed header. Sorry the comment was cryptic.

R/credentials_gce.R
```diff
@@ -27,13 +27,22 @@ credentials_gce <- function(scopes = NULL, service_account = "default", ...) {
...
+  credentials = full_credentials$access_token
+
   params <- list(
     as_header = TRUE,
     scope = scopes,
     service_account = service_account
   )
-  token <- GceToken$new(credentials = credentials, params = params)
+  # The underlying Token2 class appears to *require* an endpoint and an app, though it doesn't
+  # use them for anything in this case...
```
jennybc: This is pretty believable to me, as I've found another instance of similar:

https://github.com/r-lib/httr/issues/576

jennybc:
I'm just going to merge this because I am 99% certain it does not make things worse and quite likely make things much better. I want to release today and my process is easier if I'm doing all my pre-flight checks in `master`.

R/credentials_gce.R
wlongabaugh: Looks good to me. Thanks!

PR #101, opened 2019-08-12
Clarify credential paths

jimhester:
I feel it is more clear to use the environment variable idioms of the systems, even if there is a small amount of duplication as a result.


jennybc:
Thanks!

jennybc:
Is this even true anymore?

> with some abuse of notation

PR #107, opened 2019-08-16
Save tokens in RDS version 2 format

jimhester:
So they can be loaded in R versions prior to 3.6

### DIFF
--- a/R/oauth-cache.R
+++ b/R/oauth-cache.R
@@ -148,7 +148,7 @@ token_into_cache <- function(candidate) {
   if (is.null(cache_path)) {
     return()
   }
-  saveRDS(candidate, path(cache_path, candidate$hash()))
+  saveRDS(candidate, path(cache_path, candidate$hash()), version = 2)
 }
 
 # helpers to compare tokens based on SHORTHASH_EMAIL ------------------------


jimhester:
Currently the gargle tokens created with R 3.6 end up being compatible with only 3.6, since we don't save them in the version 2 format.

jennybc:
In general, I am being careful about serialization format. But I feel like OAuth tokens (the user ones we're caching here) should be thought of as highly perishable and not-very-portable. So I almost like the fact that the R version that made them is "baked in".

Nit: they it's a >= 3.5 thing, not 3.6, right?

Discuss 😁

MarkEdmondson1234:
Yes this happened to me and caused my whole gargle cache to be invalid upon upgrade to 3.6.  

However, if you want to keep the 3.5 cache I think its possible if you set `R_DEFAULT_SAVE_VERSION` and `R_DEFAULT_SERIALIZE_VERSION` to 2 - although I nuked all my 3.5 toke3ns before testing it.  Info from [R 3.6 release notes](https://cran.r-project.org/doc/manuals/r-release/NEWS.pdf)

So in light of that perhaps if people want to keep old tokens they can use above?

jimhester:
This wasted my time and I had to resave the token in 3.5 format to use it in CIs when testing older versions of R.

Now that I know it I will handle it myself going forward, but I figured I would save other people time debugging.

jennybc:
I've stuck with storing tokens using the system's default serialization version, because I still feel like the portability of OAuth tokens something we need to explicitly support.

But when I load the token cache, I now detect this specific problem and fix it, via cc5f33cbf1c01b965eba4e8dc1580c0e02cb295a. So I think the point of this PR has been addressed in another way.

PR #110, opened 2019-09-05
Add default scope on GCE

MarkEdmondson1234:
A tiny quality of life improvement proposal, since this function is only called when on GCE, add the default to that scope as otherwise I scratched my head a bit wondering why it wasn't working.


jennybc:
@craigcitro Is there any reason NOT to merge this?

I defer to you and @MarkEdmondson1234 because I've got no personal experience with this particular credential flow at this point.

jennybc:
Thanks!

PR #116, opened 2019-10-25
Fix default application credentials path

### LINKED ISSUE #115: Is the default path for application default credentials correct?
I'm having trouble to get application default credentials to work.

When I call the application default credentials loader directly:

```r
gargle::credentials_app_default()
```

I get the following error message:

```
Error in open.connection(con, "rb") : cannot open the connection
In addition: Warning messages:
1: In file(txt) :
  'raw = FALSE' but '/home/faculty/.config/gcloud' is not a regular file
2: In open.connection(con, "rb") :
  cannot open file '/home/faculty/.config/gcloud': it is a directory
```

gargle appears to be attempting to load `~/.config/gcloud` instead of `~/.config/gcloud/application_default_credentials.json`.

Could it be that this line

https://github.com/r-lib/gargle/blob/e54eb9d68f79a905710cc8a2a5991c89a174a3a7/R/credentials_app_default.R#L96

needs to be changed to?:

```r
pth <- path_home(".config", "gcloud", pth)
```

(Prior to this line `pth` has been assigned the value `"application_default_credentials.json"`)

acroz:
Prior to this PR, the default credentials path on non-Windows systems was determined to be:

```
~/.config/gcloud
```

rather than the correct:

```
~/.config/gcloud/applicaiton_default_credentials.json
```

This PR fixes this and adds some tests for the logic in the offending function.

This PR fixes #115, and I have validated that the change corrects the issue I had when using `bigrquery` with application default credentials.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # gargle (development version)
 
+* The application default credentials path is fixed on non-Windows platforms
+  (#115, @acroz).
+
 * `response_process()` explicitly declares the UTF-8 encoding of the content in Google API responses [tidyverse/googlesheets4#26](https://github.com/tidyverse/googlesheets4/issues/26).
 
 # gargle 0.4.0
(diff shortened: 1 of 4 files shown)


jennybc:
Thanks!

PR #131, opened 2020-02-23
Add `subject` argument to credentials_service_account()

samterfa:
Allows domain wide delegation to be used with service accounts by passing subject parameter.


jennybc:
Can you link me to any docs to educate myself better about this?

samterfa:
Domain-wide delegation is needed, for example, when using the Google Admin SDK Directory API via a service account as explained here: https://developers.google.com/admin-sdk/directory/v1/guides/delegation. In order to sign requests correctly, a "sub" parameter is passed to the JWT. httr has incorporated the "sub" parameter into its "oauth_service_token" function which passes the parameter on to the "TokenServiceAccount$new" function to sign the request correctly. All that is needed is for Gargle's "credentials_service_account" function to accept and pass on that parameter to httr's service account functions. More info on domain-wide delegation is here: https://developers.google.com/identity/protocols/OAuth2ServiceAccount#delegatingauthority. "sub" stands for "subject" so I used that as the parameter, but they could just match instead. https://google-auth.readthedocs.io/en/latest/reference/google.oauth2.service_account.html. Thanks!

samterfa:
Is this a possibility?

jennybc:
> Is this a possibility?

My lack of response just means I'm trying to get googlesheets4 ready for release and then will come here. Because, in fact, gargle must release first. So I'll be back here soon!

jennybc:
From reading Python examples and docs, it's seems like we should follow their convention and name this argument `with_subject`.

jennybc:
@samterfa Are you happy with where this ended up? I'll leave unmerged for a bit in case you have comments, but then I'll merge and move on.

jennybc:
Cc @jimhester I think this is relevant to gmailr. Remember when we were looking into whether a service account could be used with Gmail? And we concluded that it's possible but probably would not be common, since it requires delegation of domain-wide authority to the service account. This PR now makes this possible. Not sure if gmailr can use this immediately (I doubt it), but you might want to enable / document next time you do work there.

samterfa:
Looks great. I really appreciate it!

Issue #130, opened 2020-02-14
gargle fails to find credentials on Kubernetes Engine using ADC


Ozan147:
I'm trying to use `bigrquery` from within a docker container that runs on Kubernetes Engine. `credentials_app_default()` fails to authenticate and the browser option gets triggered through `credentials_user_oauth2()`. I don't experience this issue when I run the container on a GCE VM as `credentials_gce()` takes care of authentication. 

The same container runs fine on Kubernetes Engine if I replace `bigrquery` with Python's `bigquery` from `google.cloud`. As you mentioned in #78, it seems gargle's ADC strategy doesn't quite match https://cloud.google.com/docs/authentication/production:
> If the environment variable isn't set, ADC uses the default service account that Compute Engine, Kubernetes Engine, Cloud Run, App Engine, and Cloud Functions provide, for applications that run on those services.

jennybc:
Can you set this option in an affected R session and see if we can get more info about where things are going wrong?

```
options(gargle_quiet = FALSE)

## put your bigrquery operations below here
```

Ozan147:
Here is a minimal reproducible example:

`bigrquery` authenticates successfully through `gargle::credentials_gce()` when the following docker image runs on a GCE VM. However; authentication fails when the image runs on AI Platform Training, which uses Kubernetes Engine under the hood. Using Python's `google-cloud-bigquery` instead of `bigrquery` works fine in both cases.

Ideally, I'd expect `gargle::credentials_app_default()` to find app default credentials and handle this issue when the image runs on AI Platform Training so that I don't need to use a service account token.

Code:
```r
options(gargle_quiet = FALSE)

tb <- bigrquery::bq_project_query(
  x = "<project>", 
  query = "<sql>"
)

bigrquery::bq_table_download(x = tb)

print("DONE!")
```

Dockerfile:
```
FROM rocker/r-ver:3.6.1

RUN apt-get update && apt-get install -y --no-install-recommends \
    git-core \
    libssl-dev \
    libcurl4-gnutls-dev \
    && apt-get clean \
    && rm -rf /var/lib/apt/lists/

RUN install2.r -e \
    bigrquery
    
COPY script.R /

ENTRYPOINT ["Rscript", "script.R"]
```
Running the image on a GCE VM:
```
docker build -t test .
docker run --rm test

# trying token_fetch()
# trying credentials_service_account()
# Error: Argument 'txt' must be a JSON string, URL or file.
# trying credentials_app_default()
[...]

PR #139, opened 2020-05-02
Print debugging around GCE creds

jennybc:
WIP

Relates to #130


jennybc:
#130 was already closed, but #185 is also relevant re: GCE scopes

Issue #143, opened 2020-05-07
Authentication fails with a service account token


pofl:
This still worked yesterday. It _might_ be that the error is new in 0.5.0
I have put a service account token in the location at $CLOUDSDK_CONFIG
I run this code in a container

```
> options(gargle_quiet = FALSE)
> library(bigrquery)
> library(httr)
> set_config(verbose(data_out = TRUE, data_in = TRUE, info = TRUE, ssl = FALSE))
> bq_auth()
trying token_fetch()
trying credentials_service_account()
Error: Argument 'txt' must be a JSON string, URL or file.
trying credentials_app_default()
file exists at ADC path: /root/gcp-gargle/application_default_credentials.json
ADC cred type: service_account
trying credentials_service_account()
adding 'userinfo.email' scope
*  Expire in 0 ms for 6 (transfer 0x55e812033c50)
*  Expire in 10000 ms for 2 (transfer 0x55e812033c50)
*  Connection 0 seems to be dead!
*  Closing connection 0
*  17 bytes stray data read before trying h2 connection
*  Found bundle for host oauth2.googleapis.com: 0x55e81230d350 [can multiplex]
*  Re-using existing connection! (#2) with host oauth2.googleapis.com
*  Connected to oauth2.googleapis.com (172.217.23.106) port 443 (#2)
*  Using Stream ID: 3 (easy handle 0x55e812033c50)
-> POST /token HTTP/2
-> Host: oauth2.googleapis.com
-> User-Agent: libcurl/7.64.0 r-curl/4.3 httr/1.4.1
-> Accept-Encoding: deflate, gzip
-> Accept: application/json, text/xml, application/xml, */*
-> Content-Type: application/x-www-form-urlencoded
-> Content-Length: 839
->
>> grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Ajwt-bearer&assertion=...

*  We are completely uploaded and fine
<- HTTP/2 400
<- content-type: application/json; charset=UTF-8
<- vary: Origin
<- vary: X-Origin
<- vary: Referer
<- content-encoding: gzip
<- date: Thu, 07 May 2020 10:43:28 GMT
<- server: scaffolding on HTTPServer2
<- cache-control: private
<- content-length: 162
<- x-xss-protection: 0
<- x-frame-options: SAMEORIGIN
<- x-content-type-options: nosniff
[...]

Issue #111, opened 2019-10-01
Auth from Docker, using a service account, throws Bad Request (HTTP 400)


rensa:
I'm trying to package my Shiny app up in a Docker container and use a service account, rather than relying on the OAuth dance for my users. If I point my local R/gargle/googledrive setup to the service account, it appears to work fine. But if I put it in the Docker container, I get:

```
trying token_fetch()
trying credentials_service_account()
adding 'userinfo.email' scope
Error: Bad Request (HTTP 400).
trying credentials_app_default()
trying credentials_gce()
trying credentials_byo_oauth()
Error: inherits(token, "Token2.0") is not TRUE
trying credentials_user_oauth2()
Gargle2.0 initialize
attempt from: googledrive
adding 'userinfo.email' scope
loading token from the cache
no matching token in the cache
Waiting for authentication in browser...
Press Esc/Ctrl + C to abort
Please point your browser to the following url:
```

Here's the code I'm using to auth:

```r
options(gargle_quiet = FALSE)
  message(
    'Starting auth with service key @ ',
    mypackage:::app_globals$service_key)
    
  drive_auth(
    path = mypackage:::app_globals$service_key,
    scopes = c(
      'https://www.googleapis.com/auth/drive',
      'https://www.googleapis.com/auth/spreadsheets'))

  message('Auth done; looking up request sheets')
  drive_ls(ids$requests, team_drive = ids$drive) %>%
    filter(id != ids$template)
```

I'm not sure if the Bad Request error means that I'm getting the scopes wrong (although I haven't had a problem with them locally) or something else. I'm verifying when my package starts up that the file at the specified path actually exists, and I'm verifying when I run this auth function that the path is still correctly specified.

My DESCRIPTION (for the purposes of checking package versions) has: 

```
Remotes:
    tidyverse/googlesheets4@094063e,
    tidyverse/googledrive@3005625
```

jennybc:
Can you run that code interactively in the Docker container? I.e. not necessarily inside a running Shiny app.

Does the app run locally, i.e. not inside the Docker container?

I'm not sure what's going on ...

Could it have something to do with a different port landscape in the Shiny app, running inside the Docker container? Try something like `httr::with_verbose(googledrive::drive_auth())` to get more detail on what's happening.

Why not use CRAN googledrive? What version does this SHA represent?

rensa:
[...]

PR #154, opened 2020-10-05
Add system clock note to service account docs, vignette

rensa:
Adds a note to the docs (for `credentials_service_account`) and one of the vignettes about system clock issues (with particular respect to #111 and #143).

I should note that a new fix for the specific cause of #111 and #143 has been deployed to the Docker for Windows 2.4 stable release, but it's not the first fix that has gone to stable and not worked out—this issue has occurred intermittently since 2016, and WSL2 is still under active development. It seems reasonable to assume that this might continue to crop up.

Ironically, the commit datestamp for this PR is off by 16 hours due to what looks like a time zone issue 😆


vignettes/how-gargle-gets-tokens.Rmd
```diff
@@ -81,6 +81,8 @@ For R users, a service account is a great option for credentials that will be us
 
 The JSON key file must be managed securely. In particular, it should not be kept in, e.g., a GitHub repository (unless it is encrypted). The encryption strategy used by gargle and other packages is described in the article [Managing tokens securely](https://gargle.r-lib.org/articles/articles/managing-tokens-securely.html).
 
+Note that fetching a token for a service account requires a reasonably accurate system clock. This is of particular importance for users running gargle inside a Docker container, as Docker for Windows has [intermittently seen problems with clock drift](https://github.com/docker/for-win/issues/4526). If your service account token requets fail with "Bad Request" inside a container but succeed locally, check that the container's system clock is accurate.
```
pofl, line 84: Even more specifically: Docker for Windows on the WSL2 engine.

rensa: Nah, this affects both the Hyper-V and WSL2 engines. I originally encountered it with Hyper-V; I'm told it also happens on WSL2, although I haven't had the opportunity to test it (I'm on WSL2 now, but I use a work laptop from home now, so it tends to be shut down each day 😛).

pofl: Okay then wow lol.

For me it's absolutely wild. Clocks out of sync right after system boot sometimes.

rensa: Yeah, it was pretty wild to diagnose! 😬

Are there any other changes you'd like me to make here?

pofl: None that I can think of

rensa: Thanks for taking a look! 😁

jennybc:
Thanks!

PR #155, opened 2020-10-30
Look for oauth info in `web` key in client_secret.json

### LINKED ISSUE #94: oauth_app_from_json() does not read web based client JSON
The `oauth_app_from_json()` works great on client JSON for "Other" type clients, but for "Web" type clients it fails.  

The latter is needed for Shiny based auth as it includes the redirect URIs (which will be great to include later on instead of `option(googleAuthR.redirect)`)

The format is largely the same, except it starts with a "web" object and not "installed" e.g.

```json
{"web":{"client_id":"XXX.apps.googleusercontent.com",
"project_id":"blah","auth_uri":"https://accounts.google.com/o/oauth2/auth",
"token_uri":"https://accounts.google.com/o/oauth2/token",
"auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs",
"client_secret":"3nVXXXXXa4",
"redirect_uris":["http://localhost"],
"javascript_origins":["https://www.example.com","http://localhost:1221"]}}
```

Not a pressing issue until Shiny auth is in but a note for future

jcheng5:
Closes #94

To repro, create a new OAuth client ID, and for "Application type", choose "web". If you download the secrets JSON file for this, you'll see that the key is called `"web"` instead of `"installed"`.

### DIFF
--- a/R/oauth-app.R
+++ b/R/oauth-app.R
@@ -21,7 +21,12 @@ oauth_app_from_json <- function(path,
                                 appname = NULL) {
   stopifnot(is_string(path), is.null(appname) || is_string(appname))
 
-  info <- jsonlite::fromJSON(path, simplifyVector = FALSE)[["installed"]]
+  json <- jsonlite::fromJSON(path, simplifyVector = FALSE)
+  info <- json[["installed"]]
+  if (is.null(info)) {
+    # Web client credentials use this key instead
+    info <- json[["web"]]
+  }
 
   if (!all(c("client_id", "client_secret") %in% names(info))) {
     stop("Can't find 'client_id' and 'client_secret' in the JSON", call. = FALSE)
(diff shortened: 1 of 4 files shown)


R/oauth-app.R
```diff
@@ -21,7 +21,12 @@ oauth_app_from_json <- function(path,
                                 appname = NULL) {
   stopifnot(is_string(path), is.null(appname) || is_string(appname))
 
-  info <- jsonlite::fromJSON(path, simplifyVector = FALSE)[["installed"]]
+  json <- jsonlite::fromJSON(path, simplifyVector = FALSE)
+  info <- json[["installed"]]
+  if (is.null(info)) {
+    # Web client credentials use this key instead
+    info <- json[["web"]]
+  }
```
jennybc, line 29:
```suggestion
  info <- json[["installed"]] %||% json[["web"]]
```

jennybc (approved):

jennybc:
Do you have the ability to merge? If not, do you want that, for now and future?

jennybc:
Ah, I think you must, because this is coming from an internal branch.

PR #157, opened 2020-11-04
Shiny integration

jcheng5:
This might not end up being part of the gargle package, but I'm creating a draft PR here for discussion purposes.


jcheng5:
Here's some example code from googledrive that's quoted in the gargle docs:

```r
drive_token <- function() {
  if (isFALSE(.auth$auth_active)) {
    return(NULL)
  }
  if (!drive_has_token()) {
    drive_auth()
  }
  httr::config(token = .auth$cred)
}

# googledrive::
drive_has_token <- function() {
  inherits(.auth$cred, "Token2.0")
}
```

We need to somehow inject ourselves into this process and let the Shiny-session-owned auth token take precedence over the global `.auth`.

My first take looked like this:

```r
drive_token <- function() {
  ambient_token <- gargle::ambient_token()
  if (!is.null(ambient_token)) {
    return(httr::config(token = ambient_token))
  }

  if (isFALSE(.auth$auth_active)) {
    return(NULL)
  }
  if (!drive_has_token()) {
    drive_auth()
  }
  httr::config(token = .auth$cred)
}
```

but this is starting to feel like code that's too specific to be copied and pasted into every gargle-derived packages.

What if this logic lived in gargle?

```r
drive_token <- function() {
  gargle::get_token(auth_state = .auth, auth_func = drive_auth)
}
```

If we want to give users control over whether they want to respect the ambient creds from Shiny:

```r
drive_token <- function(allow_ambient = TRUE) {
  gargle::get_token(auth_state = .auth, auth_func = drive_auth, allow_ambient)
}
```

cc @jennybc @hadley

MarkEdmondson1234:
Really looking forward to this 👍 

I presume you want to do this R-server side, but I did find using the [client-side JavaScript Google auth scripts](https://github.com/MarkEdmondson1234/googleAuthR/blob/master/R/shiny-js-auth.R) ultimately the easiest way to handle things like cookie management etc. and since Shiny always needs JS no friction for the end users.  Perhaps both can be supported (client and server side) in `gargle`?.

jcheng5:
@MarkEdmondson1234 Oh, thanks for the pointer; that didn’t even occur to me, I’ve only ever used OAuth myself.

No regrets though, the server side approach is already working well and my intention is to also adapt this code for other OAuth providers like Microsoft/Azure and GitHub. I also don’t want to access cookies from JS, our commercial customers’ automated testing tools often flag non-HttpOnly cookies as a security risk.

R/shiny.R
```diff
@@ -0,0 +1,307 @@
...
+  force(oauth_app)
+  force(scopes)
+  force(welcome_ui)
+
+  httpHandler <- app$httpHandler
+  app$httpHandler <- function(req) {
+    resp <-
+      handle_oauth_callback(req, oauth_app, cookie_opts) %||%
+      handle_logged_in(req, oauth_app, httpHandler) %||%
+      handle_welcome(req, oauth_app, scopes, cookie_opts)
```
jennybc, line 14: `welcome_ui` is never used anywhere and, based purely on name, I wonder if maybe it goes here?

jennybc: Out of curiosity, why does `resp` need to be assigned?

R/shiny.R
```diff
@@ -0,0 +1,307 @@
...
+          app = oauth_app,
+          code = code,
+          redirect_uri = infer_app_url(req)
+        )
+
+        # cred has:
+        # access_token, expires_in, scope, token_type, and id_token
+        # (and possibly refresh_token)
+
+        return(shiny::httpResponse(
```
jennybc, line 57: Is there a Shiny reason to use an explicit `return()`?

R/shiny.R
```diff
@@ -0,0 +1,307 @@
...
+        ))
+      }
+    }
+  }
+}
+
+handle_logged_in <- function(req, oauth_app, httpHandler) {
+  if (!is.null(read_creds_from_cookies(req, oauth_app))) {
+    # User is already logged in, proceed
+    return(httpHandler(req))
```
jennybc, line 77: Same question as above re: `return()`

R/shiny.R
```diff
@@ -0,0 +1,307 @@
...
+  redirect_uri <- infer_app_url(req)
+  state <- sodium::bin2hex(sodium::random(32))
+  query_extra <- list(
+    access_type = "offline"
+  )
+
+  # TODO: Add email?
+
+  auth_url <- httr::oauth2.0_authorize_url(
+    endpoint = gargle_outh_endpoint(),
```
jennybc, line 91:
```suggestion
    endpoint = gargle_oauth_endpoint(),
```

R/shiny.R
```diff
@@ -0,0 +1,307 @@
...
+    oauth_app,
+    scope = paste(scopes, collapse = " "),
```
jennybc, line 93: Elsewhere, gargle always adds the `"https://www.googleapis.com/auth/userinfo.email"` scope and "normalizes" the scopes. The motivation for both is to make a better key for the token, i.e. be able to handle same person dealing with same API with 2 different Google accounts and to *not* be sensitive to the order in which scopes are provided. I assume the same considerations apply here, but I'm not very confident in that assumption.

https://github.com/r-lib/gargle/blob/1bab80cba8abb249f3c2fe3477767356bf7027ef/R/Gargle-class.R#L138

jennybc (commented):
I've read through this once and can only make minor / rather formal comments. This is definitely not the sort of code I can run in my head, since Shiny is pretty outside my wheelhouse.

As for the `drive_token()` design question, is there no way for us to put the "consider ambient token" logic inside the AuthState?

R/shiny.R
jennybc: The `userinfo.email` scope is also just plain handy, because it allows client packages to at least display some information about who we are currently auth'd as. Some APIs have a proper 'who am I?" endpoint, e.g. Drive user, whereas others do not.

R/shiny.R
jcheng5: Yep, `welcome_ui` is not implemented yet but it will go there.

I assigned `resp` for two pretty trivial reasons: 1) to get the `handle_*` calls to be all indented together, and 2) to make it clear that there's a return value here that matters (vs. calling `handle_*` invoking some side effect, which is the way many web frameworks work).

R/shiny.R
jcheng5: I often use explicit `return()` to emphasize that an expression needs to be the return value, and to be resilient to future code changes that might make this no longer the last expression. Especially in an `if` as in this case (and the one below), as I think the curly braces obscure the last value somewhat.

I know @hadley tends to only use them when necessary, we have a difference of opinion here. This being your package, I'm happy to adhere to whatever style you prefer.

R/shiny.R
jcheng5: Actually the way I have it is "correct" 😬 

I'm happy to fix the actual function in a separate PR if you like?

R/shiny.R
jcheng5: Those considerations don't apply here, but I also think there's no reason not to; the Google consent screen doesn't even mention the email and profile scopes, whether you include them or not.

jcheng5:
@jennybc Ah, yes, I think `AuthState` will work! I will need to change `cred` and `auth_active` to active properties, but other than that, it looks like it should be simple. Thanks!

R/shiny.R
jennybc: 😱

jennybc: Fixed in 848663bf7e930bf59f69956908997d782fecee2f, so you can merge/rebase

R/shiny.R
jennybc: Yeah, thus far at least, the `userinfo.email` is semi-officially regarded as a non-sensitive scope and IME has never appeared on the consent screen. I suppose that could change one day.

Can you amplify on why this scope would never be handy in a Shiny context, the way it is elsewhere?

jcheng5: The reason email doesn't matter for Shiny, is because Shiny never stores multiple tokens in one place, like the gargle cache does, and thus there never needs to be any way to identify or compare a token (i.e. no need to call `Gargle2.0$hash()`). (Shiny never uses the gargle cache, every visitor needs to auth themselves to prove they are who they say they are; and those credentials are "cached" in the cookie.)

jennybc: I think the reason to get this scope is also for the sake of the client package and, potentially, Shiny app -- so not just about labelling tokens in a store. It seems this would be needed to, for example, show the user who they are logged in as.

jcheng5: Sounds good. This'll be in the next commit I make.

jcheng5:
@jennybc There is a (not insurmountable) problem with modifying AuthState's behavior.

This is how `.auth` is declared [in googledrive](https://github.com/tidyverse/googledrive/blob/43436098863c82921ec8accf5f47e9c3aed04ad4/R/drive_auth.R#L4):

```r
.auth <- gargle::init_AuthState(...)
```

This code gets executed *at package build time*, not at package load time. So the AuthState object is sort of frozen and serialized at the time that googledrive is built. Any changes that I make now to AuthState will not apply to that object, even if googledrive is reinstalled (from binary).

What needs to happen instead is something like:

```r
.auth <- NULL

.onLoad <- function(libname, pkgname) {
  .auth <<- gargle::init_AuthState(
    package     = "googledrive",
    auth_active = TRUE
  )
}
```

Without this change, the symptom will be, if someone tries to use the Shiny OAuth integration then calls to `drive_token()` will either return nothing, or return a token that is already activated on the `googledrive::.auth` object.

My proposal is:

1. I'll file a PR on all gargle-derived packages making this change, and change gargle docs to recommend the new pattern.
2. For all such packages, take note of the minimum version that will work with the new Shiny OAuth.
3. While Shiny OAuth is running, register a `token_fetch` function that detects calls by obsolete versions, and emit a message instructing them to upgrade to the correct version.

jennybc:
> modifying AuthState's behavior

Sounds good.

> So the AuthState object is sort of frozen and serialized at the time that googledrive is built.

Well, the *logic and capabilities* of AuthState are determined at build time, but not the actual tokens that it manages. Right? Tokens persist between sessions via the cache, not via `.auth`.

But, yes, client packages will need to change to work with Shiny-capable gargle.

Do you agree with this take?

jennybc:
One thing we haven't discussed is people who want to use a cached OAuth token or a service account token in a Shiny app, i.e. Shiny apps that *don't* want to auth as the user of the app. Maybe this already "just works" or, more accurately, "still works". Because this is the one thing that was already working with existing gargle and client packages.

jcheng5:
> Do you agree with this take?

Yep, 100%.

> One thing we haven't discussed is people who want to use a cached OAuth token or a service account token in a Shiny app, i.e. Shiny apps that don't want to auth as the user of the app. Maybe this already "just works" or, more accurately, "still works". Because this is the one thing that was already working with existing gargle and client packages.

This should still work exactly as before; all of the functionality I'm adding is opt-in, by piping your Shiny app object to `require_oauth`:

```r
shinyApp(ui, server) %>% require_oauth(oauth_app, scopes)
```

What I am precluding is the possibility of using a cached token or service account *while* also using `require_oauth` (i.e. some calls use a cached/service token, some use the visitor's token). This isn't an inherent limitation of the approach, but I thought it was the right place to start, given how disastrous it'd be to accidentally let random visitors use a cached credential, and how difficult it is to detect this when you as a developer are probably testing by logging in using your own credentials that are also cached!

My approach also does nothing for apps that want to make login optional, or to progressively reveal functionality if the user logs in from a button within an app. We'll need to make it clear in the docs exactly what is supported.

jcheng5:
@hadley @jennybc I think the time has come to decide where we want this code to ship.

* In {gargle}? Then {base64enc} becomes a dependency and {shiny} (and some packages shiny already depends on) becomes a Suggests or Enhances.
* In {shiny}? Feels too domain-specific, especially given the explicit tie to Google
* In a new {shinygargle} package
* In a new {shinyoauth} package - but I don't know whether/when I'd get around to adapting this code for non-Google scenarios

jennybc:
A few notes from me playing with the PR

---

```
> checking dependencies in R code ... NOTE
  Missing or unexported object: ‘shiny::httpResponse’
```

Do we need a minimum version on shiny? In fact, should it be in `Remotes:`, i.e. we need a dev version?

---

```
> checking Rd cross-references ... WARNING
  Missing link or links in documentation object 'require_oauth.Rd':
    ‘google_signin_button’
 ```

**Comment by jcheng5** [comment]
Oh yes so sorry. Needs shiny master.

File: R/AuthState-class.R [hunk 6]
```diff
@@ -86,89 +86,147 @@ init_AuthState <- function(package = NA_character_,
...
+      self$api_key     <- api_key
+      self$auth_active <- auth_active
+      self$cred        <- cred
+      self
+    },
+    #' @description Print an AuthState
+    #' @param ... Not used.
+    print = function(...) {
+      withr::local_options(list(gargle_quiet = FALSE))
+      ui_line("<AuthState (via gargle--with Joe's changes)>")
```
jcheng5, line 127: Whoops

PR #158, opened 2020-11-22
add function suggestion?

maelle:
although this might not be relevant for secrets in gargle? thanks a ton for this vignette in any case


jennybc:
Sure, it's fine to mention, if it's helped you.

PR #162, opened 2021-01-08
Remove potential warnings about utf8 encoding

michaelquinn32:
Not all systems recognize "utf8" as a valid encoding. "UTF-8" is probably better.
Ref: https://stat.ethz.ch/R-manual/R-devel/library/base/html/Encoding.html


jennybc:
Thanks!

PR #172, opened 2021-03-31
Introduce special handling for 'deleted client'

### LINKED ISSUE #168: Can I catch and "reframe" error thrown when we disable an oauth app?
When we roll the tidyverse oauth app and actually disable the previous one, can I catch the error users will see if they haven't upgraded client package + gargle and supplement it with a probably diagnosis and suggested solution?

Obviously this won't help this time, but it could help next time.

jennybc:
Closes #168

### DIFF
--- a/R/Gargle-class.R
+++ b/R/Gargle-class.R
@@ -214,8 +214,10 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
     )
     if (is.null(cred)) {
       token_remove_from_cache(self)
-      # TODO: return(NULL) here? why do we return the current, invalid,
-      # unrefreshed token?
+      # TODO: why do we return the current, invalid, unrefreshed token?
+      # at the very least, let's clear the refresh_token, to prevent
+      # future refresh attempts
+      self$credentials$refresh_token <- NULL
     } else {
       self$credentials <- cred
       self$cache()
--- a/R/Gargle-class.R
+++ b/R/Gargle-class.R
@@ -208,7 +208,7 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
   },
   #' @description (Attempt to) refresh a Gargle2.0 token
   refresh = function() {
-    cred <- get("refresh_oauth2.0", asNamespace("httr"))(
+    cred <- refresh_oauth2.0(
       self$endpoint, self$app,
       self$credentials, self$params$user_params, self$params$use_basic_auth
     )
--- a/R/Gargle-class.R
+++ b/R/Gargle-class.R
@@ -214,6 +214,8 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
     )
     if (is.null(cred)) {
       token_remove_from_cache(self)
+      # TODO: return(NULL) here? why do we return the current, invalid,
+      # unrefreshed token?
     } else {
       self$credentials <- cred
       self$cache()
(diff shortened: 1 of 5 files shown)


R/Gargle-class.R
```diff
@@ -208,12 +208,14 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
   },
   #' @description (Attempt to) refresh a Gargle2.0 token
   refresh = function() {
-    cred <- get("refresh_oauth2.0", asNamespace("httr"))(
-      self$endpoint, self$app,
-      self$credentials, self$params$user_params, self$params$use_basic_auth
+    cred <- refresh_oauth2.0(
```
jennybc, line 211: I was already doing something janky with this method in order to use a `Gargle2.0`-specific cache-removal procedure , so it feels fine/good to take the next step and fully implement `refresh_oauth2.0()` in gargle.

R/Gargle-class.R
```diff
@@ -208,12 +208,14 @@ Gargle2.0 <- R6::R6Class("Gargle2.0", inherit = httr::Token2.0, list(
...
+      self$endpoint, self$app, self$credentials,
+      package = self$package
     )
     if (is.null(cred)) {
       token_remove_from_cache(self)
+      # TODO: return(NULL) here? why do we return the current, invalid,
+      # unrefreshed token?
```
jennybc, line 218: The more I think about this, I think, when a token fails to refresh, we should clear its refresh token. And possibly do more to indicate that it's invalid and why?

Long-term, httr2-wise, I think a token needs a way to communicate its viability back to a client package that may be managing it in memory.

R/oauth-cache.R
```diff
@@ -166,6 +166,8 @@ cache_clean <- function(cache, pattern = "-calliope$") {
   TRUE
 }
 
+gargle_legacy_app_pattern <- function() "-calliope$"
```
jennybc, line 169: I assume this would evolve over time, as we roll the apps.

R/oauth-refresh.R
```diff
@@ -0,0 +1,113 @@
+# this file has its origins in oauth-refresh.R and oauth-error.R in httr
+# I want to introduce behaviour to error informatively for a deleted OAuth app
+
+# Refresh an OAuth 2.0 credential.
+#
+# Refreshes the given token, and returns a new credential with a
+# valid access_token. Based on:
+# https://developers.google.com/identity/protocols/oauth2/native-app#offline
+refresh_oauth2.0 <- function(endpoint, app, credentials, package = NULL) {
```
jennybc, line 9: Pretty faithful copy of `httr::refresh_oauth2.0()`, but with some simplifications re: things that can't happen with Google auth + a relaxation of rigid lack-of-refresh error types.

R/oauth-refresh.R
```diff
@@ -0,0 +1,113 @@
...
+
+# This implements error checking according to the OAuth2.0
+# specification: https://tools.ietf.org/html/rfc6749#section-5.2
+find_oauth2.0_error <- function(response) {
+  if (!httr::status_code(response) %in% oauth2.0_error_codes) {
+    return(NULL)
+  }
+
+  content <- httr::content(response)
+  if (is.null(content$error)) {
```
jennybc, line 48: This is more lax than httr, which checks against a fixed set of strings that, apparently, Google violates.

https://github.com/r-lib/httr/commit/ce8439f2d571e91bf846b43d5db3d89610568fa1

jennybc:
I'm happy to receive feedback. But I'm also making and reviewing PR more as a way to document some analysis that might be useful for httr2. I'm punting on a few things here in a way that feels fairly harmless / way better than doing nothing.

R/Gargle-class.R
hadley: Yeah, seems like this should error as well as removing from the cache?

R/oauth-refresh.R
hadley: That was probably written when I mistakenly believed that most APIs followed the OAuth spec.

hadley (approved):

PR #178, opened 2021-06-02
Add credential fetching for workload identity federation

### LINKED ISSUE #176: bq_auth fails to authenticate using identity federation configuration file
Hi Team,

I am trying to access BigQuery from r-studio (setup on EC2) by implementing identity federation(reference https://cloud.google.com/iam/docs/access-resources-aws).

After successfully configuring workload identity pool setup and binding AWS role, I was able to generate the below JSON configuration file using the command `gcloud iam workload-identity-pools create-cred-config`

```
{
  "subject_token_type": "urn:ietf:params:aws:token-type:aws4_request",
  "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/<service-account>@<project-name>.iam.gserviceaccount.com:generateAccessToken",
  "type": "external_account",
  "credential_source": {
    "environment_id": "aws1",
    "url": "http://169.254.169.254/latest/meta-data/iam/security-credentials",
    "regional_cred_verification_url": "https://sts.{region}.amazonaws.com?Action=GetCallerIdentity&Version=2011-06-15",
    "region_url": "http://169.254.169.254/latest/meta-data/placement/availability-zone"
  },
  "token_url": "https://sts.googleapis.com/v1/token",
  "audience": "//iam.googleapis.com/projects/<projectid>/locations/global/workloadIdentityPools/<workload-identity-pool-id>/providers/<aws-workload-pool-id>"
}
```

I have set the environment variable GOOGLE_APPLICATION_CREDENTIALS to the file path and also provided an absolute path to the configuration file path as shown below.

```r
library(bigrquery) 
#bq_auth(path="/home/ubuntu/poc-bigquery-federation-creds.json")
bigrquery::list_tables('<project>', '<bq-dataset>')
```

**Actual Behaviour when using the environment variable on R:**
  Error: Can't get Google credentials.
  Are you running big query in a non-interactive session? Consider:
    * Call `bq_auth()` directly with all necessary specifics.

**Actual Behaviour when set the absolute path to the configuration file:**
The bigrquery package is requesting access to your Google account
  Select a pre-authorised account or enter '0' to obtain a new token
  Press Esc/Ctrl + C to cancel

**Expected Behaviour on R:**
  1. authenticate bq_auth automatically using federation configuration file provided
  2. bigrquery able to list tables 
  
I would like to confirm that with the above configuration file, google python-bigquery is able to list the tables on big query. Kindly let me know if there is a way to authenticate in R using the above federation configuration file.

Please let me know if you need more information from me.

jennybc:
Closes #176
(Well, closes it for AWS. I will address other providers or generalize the AWS setup when and if actual users have such a need.)

Open question: is it inevitable that this only works when `scopes = "https://www.googleapis.com/auth/cloud-platform"`?

---

~Currently this PR depends on a patched version of the aws.signature package. I need to figure out how likely they are to adapt to what I want/need and then release to CRAN in a timely manner. I will probably need to implement what I need myself, probably by calling aws.signature in a different way (using low-level functions instead of high-level).~

~I could even implement everything I need from aws.ec2metadata and aws.signature myself, but I'd rather not take on so much responsibility for maintaining AWS tooling.~

*I have done what's necessary to work with the CRAN version of aws.signature. I will also look into whether PAWS could be used instead.*


jennybc:
On AWS, with workload identity federation configured correctly on AWS and GCP, I now see:

``` r
# makes the Application Default Credentials search strategy find the config
Sys.setenv(GOOGLE_APPLICATION_CREDENTIALS = "~/poc-bigquery-federation.json")

library(bigrquery)

# for some reason, I have to specify this EXACT scope and ONLY this scope
bq_auth(scopes = "https://www.googleapis.com/auth/cloud-platform")

project_name <- "vinay-playground-01"
bq_project_datasets(project_name)
#> [[1]]
#> <bq_dataset> vinay-playground-01.rbnc_training_dataset_001
#> 
#> [[2]]
#> <bq_dataset> vinay-playground-01.rbnc_training_ds
```

<sup>Created on 2021-06-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

PR #179, opened 2021-06-08
Support domain-only email specification

### LINKED ISSUE #118: Email specification at the level of domain
Idea from @Hadley based on his Google analytics work:

Matching email and pre-approving token discovery in a `PKG_auth()` function could be based on domain, e.g. `domain = "rstudio.com". This is intermediate between `email = TRUE` and having to put a specific email into a script.

This way different people at the same company could share a script, without incessantly editing the `email`.

jennybc:
Closes #118 

I think the main / only thing to discuss here is whether domain-only email specification looks like:

* `email = "example.com"` (what I've currently done, based on @hadley's input via slack)
* `email = "@example.com"`
* `email = "*@example.com"`

I think I have a weak preference for the latter two options.

The inclusion of `@` catches my eye and helps me see "hey, yo, just the domain is being specified!".

The inclusion of `*` hooks into my prior expectations around globbing.

### DIFF
--- a/vignettes/non-interactive-auth.Rmd
+++ b/vignettes/non-interactive-auth.Rmd
@@ -50,14 +50,14 @@ TL;DR is that you need to successfully authenticate *once* in an interactive ses
 ```{r}
 # Approach #1: use an option.
 # Either specify the user:
-options(gargle_oauth_email = "jenny@example.org")
+options(gargle_oauth_email = "jenny@example.com")
 # Or, if you don't use multiple Google identities, you can be more vague:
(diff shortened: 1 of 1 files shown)


hadley:
I didn't realise that `*` was already an option. In that case, I think I prefer `*@rstudio.com`.

jennybc:
> I didn't realise that `*` was already an option ...

Well, that is purely an internal matter. It's not part of the user interface.

But I am still thinking `*@rstudio.com` is the way to go.

PR #185, opened 2021-07-02
Drop proactive scope validation in credentials_gce.

### LINKED ISSUE #161: Why are allowed scopes limited for credentials_gce?
Thank you so much for your work on this package! I had struggled to pass in credentials to my publicly available Plumber API application running on Google Cloud Run without a bunch of gcloud commands, yaml files, or just baking the credentials into the application (not a good idea, I know). [credentials_gce()](https://github.com/r-lib/gargle/blob/master/R/credentials_gce.R) allows me to grab secrets from Google Secrets Manager super easily with a quick API call during the build.

I'm wondering if there's a compelling reason that credentials_gce() limits the scopes for which a token can be fetched to Google Cloud and Big Query. The Google Secrets API call I use works because the scope needed is just "https://www.googleapis.com/auth/cloud-platform" which is on the blessed list of scopes.

```
# We add a special case for the cloud-platform -> bigquery scope implication.
  if ("https://www.googleapis.com/auth/cloud-platform" %in% instance_scopes) {
    instance_scopes <- c(
      "https://www.googleapis.com/auth/bigquery",
      instance_scopes
    )
  }
  if (!all(scopes %in% instance_scopes)) {
    return(NULL)
  }
```

By adding  "https://www.googleapis.com/auth/pubsub" to the instance_scopes allowed in credentials_gce(), and giving my default compute service account in Cloud Run the Pub/Sub Publisher role, I was able to run PubSub API calls without passing in credentials to the application. As I understand it, something like this is Google's preferred method of passing in Google Cloud API credentials to an application based on [this page](https://cloud.google.com/docs/authentication/production): "If your application runs inside a Google Cloud environment that has a default service account, your application can retrieve the service account credentials to call Google Cloud APIs. Such environments include Compute Engine, Google Kubernetes Engine, App Engine, Cloud Run, and Cloud Functions. We recommend using this strategy because it is more convenient and secure than manually passing credentials."

I do understand that I could generate a json file credential, upload it to Google Secrets Manager, then load it via the same API calls I'm making, but it's less convenient. Maybe it's better practice though.

craigcitro:
Why? Avoid confusion like #161.

When adding support for fetching credentials on a GCE VM, I had originally
added a "guard rail": if the user was requesting scope(s) which weren't
included in the list of scopes the service account was granted, I'd return
`NULL`, letting the credential fetch fall back to another method (eg user
creds).

However, this can lead to confusion: most Google API calls are willing to
accept various scopes, and some scopes imply other scopes (eg an API asking
for a BQ scope will happily accept the cloud-platform scope).

This change removes any scope validation in credentials_gce. On the upside, it
should make it easier for users to use gargle from a GCE VM that *has* the
required scopes. Unfortunately, it does allow one new bit of confusion: a user
invoking gargle on a GCE VM *without* the scopes needed for an API would
previously fall through to fetching user creds; they'll now need to do so
explicitly.

Fixes #161.

### DIFF
--- a/R/credentials_gce.R
+++ b/R/credentials_gce.R
@@ -17,18 +17,7 @@
 credentials_gce <- function(scopes = "https://www.googleapis.com/auth/cloud-platform",
                             service_account = "default", ...) {
   gargle_debug("trying {.fun credentials_gce}")
-  if (!detect_gce() || is.null(scopes)) {
-    return(NULL)
-  }
-  instance_scopes <- get_instance_scopes(service_account = service_account)
-  # We add a special case for the cloud-platform -> bigquery scope implication.
-  if ("https://www.googleapis.com/auth/cloud-platform" %in% instance_scopes) {
-    instance_scopes <- c(
-      "https://www.googleapis.com/auth/bigquery",
-      instance_scopes
-    )
-  }
-  if (!all(scopes %in% instance_scopes)) {
+  if (!detect_gce()) {
     return(NULL)
   }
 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks for providing the context of the original decision and why we're OK with removing these checks. Sounds good to me.

A gargle release was in flight (actually completed) before this PR came in, so, for better or worse, this will get a chance to be road-tested in the dev version for a little while. Hopefully those who care, such as the participants in #161, will install a dev version and accumulate some usage with the guard rail removed.

jmcarp:
I think this will fix an issue I've had trying to use google drive with gce auth. Looking forward to the next release!

PR #195, opened 2021-11-02
Increasing timeout from 0.8 to 2

downloaderfan:
credentials_gce() function was failing since default timeout was too low. Increasing it from 0.8 to 2 fixed the issue.

### DIFF
--- a/R/credentials_gce.R
+++ b/R/credentials_gce.R
@@ -103,7 +103,7 @@ gce_metadata_request <- function(path, stop_on_error = TRUE) {
     path <- substring(path, 2)
   }
   url <- paste0(root_url, "computeMetadata/v1/", path)
-  timeout <- getOption("gargle.gce.timeout", default = 2)
+  timeout <- getOption("gargle.gce.timeout", default = ifelse(!detect_gce(), 0.8, 2))
   response <- try({
     httr::with_config(httr::timeout(timeout), {
       httr::GET(url, httr::add_headers("Metadata-Flavor" = "Google"))
--- a/R/credentials_gce.R
+++ b/R/credentials_gce.R
@@ -103,7 +103,7 @@ gce_metadata_request <- function(path, stop_on_error = TRUE) {
     path <- substring(path, 2)
   }
   url <- paste0(root_url, "computeMetadata/v1/", path)
-  timeout <- getOption("gargle.gce.timeout", default = 0.8)
+  timeout <- getOption("gargle.gce.timeout", default = 2)
   response <- try({
     httr::with_config(httr::timeout(timeout), {
       httr::GET(url, httr::add_headers("Metadata-Flavor" = "Google"))


craigcitro:
To confirm: did you run into the timeout on a GCE instance running the usual metadata server?

In particular, there's a tension here: you want a long timeout when you're *on* GCE, but you want a short timeout when you're *not on* GCE and you're stuck waiting for this request to timeout. (Whether the request will immediately fail or timeout depends on your networking config, sometimes including factors outside your control.)

downloaderfan:
@craigcitro Yup, I was simply running credentials_gce() function to auth into BQ using default compute engine account & found that the below line was failing

```response <- try({
  httr::with_config(httr::timeout(timeout), {
    httr::GET('http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token', httr::add_headers("Metadata-Flavor" = "Google"))
  })
}, silent = TRUE)```

It is the default metadata server request I believe. On increasing timeout from 0.8 to 2, it started working.

The particular GET request doesn't seem to work outside of GCP, it's specific to the VM.

**Comment by craigcitro** [comment]
We use the same request to detect that we're on GCE (eg https://github.com/r-lib/gargle/blob/4732f8c900b6c2733d907de0fe4023be70a4c3da/R/credentials_gce.R#L129-L132).

I suspect raising the timeout is worthwhile, but it's worth noting that a large timeout may lead to user-visible slowness for some users when checking whether or not they're on GCE.

**Comment by jennybc** [comment]
We could make this timeout configurable via an env var, with the default short, suitable for people who are not on GCE (the vast majority of gargle usage, I suspect). But GCE users could set it to something much longer.

**Comment by downloaderfan** [comment]
@craigcitro I understand your point, as @jennybc has said, it would be great if we can make it configurable with the default value being the same as it is now so that it doesn't affect existing users. Given the current state, it is too low such that it is timing out before the metadata server can return a legitimate request and since I'm running this on Google Compute Engine, it's a server to server connection so I don't think network speed should be the problem. I believe it would be worthwhile to make it configurable.

**Comment by craigcitro** [comment]
👍 on configurability.

Actually, as a related tweak: in `detect_gce`, if we detect that we *are* on GCE, maybe we should then increase the timeout, since we know it's worth waiting? This would fix your situation here (credential fetch itself was slow) without any extra work on your part, and be 0 cost for non-GCE users.

**Comment by jennybc** [comment]
I feel like there is some risk of going in circles here.

Early on, we call `detect_gce()`, which calls `gce_metadata_request()`, which is where we're talking about a longer timeout. But any call to `token_fetch()` can potentially hit this code.

I understand that we might increase the timeout once we know we're on GCE. But it feels like the only way to know we're on GCE also involves a metadata request. So for that first call, only a timeout-increase-via-config would really help.

Do I have this right?

**Comment by jennybc** [comment]
Oh maybe this is what I was missing:

`gce_metadata_request("", stop_on_error = FALSE)` is fast. This is what's in `detect_gce()`.

`gce_metadata_request(glue("instance/service-accounts/{service_account}/token"))` is slower. This is what's in `fetch_gce_access_token()`.

**Comment by jennybc** [comment]
So I think I'm on board with this suggestion from @craigcitro now:

> if we detect that we are on GCE, maybe we should then increase the timeout, since we know it's worth waiting? This would fix your situation here (credential fetch itself was slow) without any extra work on your part, and be 0 cost for non-GCE users.

**Comment by downloaderfan** [comment]
> 👍 on configurability.
> 
> Actually, as a related tweak: in `detect_gce`, if we detect that we _are_ on GCE, maybe we should then increase the timeout, since we know it's worth waiting? This would fix your situation here (credential fetch itself was slow) without any extra work on your part, and be 0 cost for non-GCE users.

Great idea!!! Yes, that would solve this problem without any additional configuration, such that I can discard any separate code for auth in all my scripts.

**Comment by downloaderfan** [comment]
We can replace the line

```timeout <- getOption("gargle.gce.timeout", default = 2)```

with

```timeout <- getOption("gargle.gce.timeout", default = ifelse(!detect_gce(), 0.8, 2))```

Let me know if should close this pull request and create a new one for the same.

File: R/credentials_gce.R [hunk 1]
```diff
@@ -103,7 +103,7 @@ gce_metadata_request <- function(path, stop_on_error = TRUE) {
     path <- substring(path, 2)
   }
   url <- paste0(root_url, "computeMetadata/v1/", path)
-  timeout <- getOption("gargle.gce.timeout", default = 2)
+  timeout <- getOption("gargle.gce.timeout", default = ifelse(!detect_gce(), 0.8, 2))
```
jennybc, line 106: This would seem to create a recursive call to `gce_metadata_request()`.

Now that we seem to be converging on the right solution, I plan to make a PR that implements it.

downloaderfan: @jennybc Alright, thanks :)

PR #180, opened 2021-06-10
Change URL for refresh

jennybc:
Motivated by #147 

I am very skeptical that this URL is the problem, but it is easy enough to do so, in case someone wants to see if this helps.

PR #196, opened 2021-11-03
Update auth URIs and use ours (vs httr's) when refreshing ADC creds

### LINKED ISSUE #147: bq_auth() relying on credentials_app_default() cannot authenticate offline with ADC
`bigrquery` version 1.3.1 cannot authenticate using Application Default Credentials on Compute Engine without Internet access in GCP (instance still has access to "internal" Google Cloud Platform APIs). 

```r
bq_auth(use_oob=TRUE)

trying token_fetch()
trying credentials_service_account()
Error: Argument 'txt' must be a JSON string, URL or file.
trying credentials_app_default()
file exists at ADC path: /home/jlpuser/.config/gcloud/application_default_credentials.json
ADC cred type: authorized_user
Error: Timeout was reached: [accounts.google.com] Connection timed out after 10001 milliseconds
trying credentials_gce()
trying credentials_byo_oauth()
Error: inherits(token, "Token2.0") is not TRUE
trying credentials_user_oauth2()
Gargle2.0 initialize
attempt from: bigrquery
adding 'userinfo.email' scope
loading token from the cache
no matching token in the cache
initiating new token

Enter authorization code: 
```

The above offline ADC authentication flow works with the Python libraries on the same type of machines.

We believe that the [this line](https://github.com/r-lib/gargle/blob/master/R/credentials_app_default.R#L65) in `gargle` should look something like `endpoint <- "https://oauth2.googleapis.com/token"` as the application default credentials file includes a refresh token as per the "offline" access_type described [here](https://developers.google.com/identity/protocols/oauth2/web-server#offline). Passing the resulting token in should prevent the library from trying to acquire it's own.

I have opened an issue first in the `bigrquery` repo by mistake. [Here is the link to it](https://github.com/r-dbi/bigrquery/issues/392).

jennybc:
Fixes #147, closes #180

`credentials_app_default()` should have always been using `gargle_oauth_endpoint()`, at least once `gargle_oauth_endpoint()` was created.

And it's pretty clear that httr's google endpoint URIs are out of date.

See:

https://github.com/googleapis/oauth2client/issues/742

https://developers.google.com/identity/protocols/oauth2/native-app#step-2:-send-a-request-to-googles-oauth-2.0-server

https://developers.google.com/identity/protocols/oauth2/native-app#exchange-authorization-code

https://developers.google.com/identity/protocols/oauth2/native-app#tokenrevoke

https://developers.google.com/identity/protocols/oauth2/service-account#authorizingrequests

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # gargle (development version)
 
+We have switched to newer oauth2.googleapis.com-based OAuth2 URIs, moving away from the accounts.google.com and googleapis.com/oauth2 equivalents.
+
 `credentials_gce()` no longer validates the requested scopes against instance scopes.
 In practice, it's easy for this check to be more of a nuisance than a help (#161, #185 @craigcitro).
 
--- a/R/credentials_app_default.R
+++ b/R/credentials_app_default.R
@@ -71,11 +71,10 @@ credentials_app_default <- function(scopes = NULL, ..., subject = NULL) {
       return(NULL)
     }
     gargle_debug("ADC cred type: {.val authorized_user}")
-    endpoint <- httr::oauth_endpoints("google")
     app <- httr::oauth_app("google", info$client_id, secret = info$client_secret)
     scope <- "https://www.googleapis.com/auth/cloud.platform"
     token <- httr::Token2.0$new(
-        endpoint = endpoint,
+        endpoint = gargle_oauth_endpoint(),
         app = app,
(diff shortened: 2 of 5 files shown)


craigcitro (commented):
LGTM

(I realize I merged before I noticed I had to separately hit a button to LG. 🤦 )

jennybc:
Thanks for looking over this and for the brisk process 😁

PR #213, opened 2022-09-09
Prototype OOB-style flow for Workbench

jcheng5:
Try adding `https://codepen.io/jcheng5/full/QWrNgPv` to your Google app as a Redirect URI, then:

```r
library(googledrive)
library(gargle)

the_app <- gargle::oauth_app_from_json("app.json")

# tell googledrive to auth with this app
drive_auth_configure(app = the_app)

# confirm the configuration has worked
drive_oauth_app()

# force a new user OAuth flow
# this fails because of OOB being shut down
drive_auth(email = NA)
```

After the Google part of the auth flow completes, you should end up at a page that has:

image

Lots more to talk about, but this shows that the approach should work.


jennybc:
Important note: the oauth client must be of type "web application", not "desktop app" (which is what we've used with gargle up 'til now). In particular, this is what makes it possible to specify redirect URI(s).

PR #214, opened 2022-09-29
Pseudo-OOB auth flow

### LINKED ISSUE #202: Out-of-bound flow is being deprecated
https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html?m=1#disallowed-oob

> OAuth [out-of-band](https://developers.google.com/identity/protocols/oauth2/native-app#manual-copypaste) (OOB) is a legacy flow developed to support native clients which do not have a redirect URI like web apps to accept the credentials after a user approves an OAuth consent request. The OOB flow poses a remote phishing risk and clients must migrate to an alternative method to protect against this vulnerability. New clients will be unable to use this flow starting on Feb 28, 2022.

> [Key dates for compliance](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html?m=1#dates-oob)
> * Feb 28, 2022 - new OAuth usage will be blocked for the OOB flow
> * Sep 5, 2022 - a user-facing warning message may be displayed to non-compliant OAuth requests
> * Oct 3, 2022 - the OOB flow is deprecated for existing clients

jennybc:
Closes #202 

Pseudo-OOB will "just work" for users on RStudio Server/Workbench/Cloud for packages that use the built-in client (actually, clients, with an "s", now that we've added a web client), which is googledrive, googlesheets4, bigrquery.

For folks who want to use pseudo-OOB with their own client, it would look something like this, with a googledrive example:

``` r
library(googledrive)

# configure a custom OAuth client (or "app")
# VERY IMPORTANT: the client must be of type "web application"
# the redirect URI must point at a static "code exchanger" web page
# with content seen here in inst/pseudo-oob/google-callback.html
drive_auth_configure(
  path = "path/to/oauth/client/json/downloaded/from/developer/console/foo.json"
)

drive_auth(email = NA, use_oob = TRUE)
```


R/oauth-init.R
```diff
@@ -0,0 +1,180 @@
...
+      # So we must explicitly ask for re-consent.
+      query_authorize_extra[["prompt"]] <- "consent"
+
+      state <- csrf_token()
+    } else { # conventional oob
+      state <- NULL
+    }
+  } else {
+    redirect_uri <- app$redirect_uri
+    state <- csrf_token()
```
jennybc, line 75: This replaces a call to `httr:::nonce()`

Issue #222, opened 2022-11-12
Clarify Google Cloud Run support


jnolis:
Hi everyone! I'm trying to connect to a Google Cloud Storage bucket from a Docker container running in Google Cloud Run. It _feels_ like this would already be supported in gargle, but I don't see any documentation either way. This leads to my two proposals:

1. Can someone (and this could easily be me), make an article in the documentation on how one should get credentials when running Docker containers within GCP? That could be either Google Cloud Run, kubernetes, or anything like that. It looks like a [previous issue](https://github.com/r-lib/gargle/issues/130) discussed it, but I couldn't understand the outcome.
2. If gargle doesn't have the functionality to get credentials from Docker containers in GCP, could that be added? I do not understand the complexities of GCP enough to know if that's a small or big request.

Thanks! Love the package.

jennybc:
I think I'll need some help (or time or, TBH, both) to get my head around what exactly needs to happen here. This is a usage scenario that just doesn't come up for me, so it's much harder for me to get oriented.

MarkEdmondson1234:
I now have some example R code I put in https://github.com/MarkEdmondson1234/googleCloudRunner/commit/a435d801b9eb10eb2acd8977a4c38d890e79795c

### Deploy to Cloud Run and reusing the default authentication

When you deploy to Cloud Run, you can choose which service key the Cloud Run service will run under, the default being the GCE default service key.  This means you can authenticate using this key without needing to upload your own service JSON file.  An example is available in the example Cloud Run app included with the package, deployable via `cr_deploy_plumber(system.file("example", package = "googleCloudRunner"))`

The relevant R code is shown below, which lets you list a Google Cloud Storage bucket within the same GCP project, reusing the default authentication.

```r
#' List a Google Cloud Storage bucket as an auth example
#' @get /gcs_list
#' @param bucket the bucket to list.  Must be authenticated for this Cloud Run service account
function(bucket=NULL){
  if(is.null(bucket)){
    return("No bucket specified in URL parameter e.g ?bucket=my-bucket")
  }

  library(googleCloudStorageR)

  auth <- gargle::credentials_gce()
  if(is.null(auth)){
    return("Could not authenticate")
  }

  message("Authenticated with service token")

  # put it into googleCloudStorageR auth
  gcs_auth(token = auth)

  gcs_list_objects(bucket)

}
```

[...]

PR #223, opened 2022-11-21
add workload identity docs

### LINKED ISSUE #197: How to run Workload Identity on GKE
Hi,

I'm using multiple `google*Rverse` R packages, and I would like to authenticate via [Workload Identity on GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity). (Right now I do it via service account json-s.)

Is this currently possible via `gargle`? I was looking at the [documentation page](https://gargle.r-lib.org/reference/credentials_external_account.html), but I gather that it will only work on AWS. If that's not the case, would it be possible to update the documentation with a more descriptive example of the steps needed to follow?

If it's not supported, would it be possible to implement it?

Thank you 🙏 
Tamas

MarkEdmondson1234:
Closes #197, relates to #222


vignettes/non-interactive-auth.Rmd
```diff
@@ -122,6 +122,50 @@ options(gargle_verbosity = "debug")
...
+
+Using Workload Identity is the recommended way to do authentication on GKE and other places if possible since it involves not downloading service keys which is a potential security risk.
+
+1. Following the [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) you create a service account as normal and give it permissions and scopes needed to say upload to BigQuery, as you would before. eg. `my-service-key@my-project.iam.gserviceaccount.com` with `https://www.googleapis.com/auth/bigquery` scopes
```
jennybc, line 131:
```suggestion
1. Following the [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) you create a service account as normal and give it permissions and scopes needed to, for example, upload to BigQuery, as you would before. eg. `my-service-key@my-project.iam.gserviceaccount.com` with `https://www.googleapis.com/auth/bigquery` scopes
```

vignettes/non-interactive-auth.Rmd
```diff
@@ -122,6 +122,50 @@ options(gargle_verbosity = "debug")
...
+gcloud iam service-accounts add-iam-policy-binding my-service-key@my-project.iam.gserviceaccount.com \
+    --role roles/iam.workloadIdentityUser \
+    --member "serviceAccount:my-project.svc.id.goog[my-namespace/bq-service-account]"
+# Annotate k8s SA
+kubectl annotate serviceaccount bq-service-account \
+    --namespace my-namespace \
+    iam.gke.io/gcp-service-account=my-service-key@my-project.iam.gserviceaccount.com
+```
+
+This key will now be available to add to pods within the cluster.  For Airflow, you can pass them in using the `GKEPodOperator(...., namespace='my-namespace', service_account_name='bq-service-account')`
```
jennybc, line 155: Is the call to `GKEPodOperator()` R code? If so, can you add namespace qualification, e.g. `pkg::fcn()`?

vignettes/non-interactive-auth.Rmd
```diff
@@ -122,6 +122,50 @@ options(gargle_verbosity = "debug")
...
+
+5. When calling the `gargle::gce_credentials()` within R, you need first make sure its using the right endpoint (`options(gargle.gce.use_ip = TRUE)`) and then call the service email that is not "default".  `gargle:::list_service_accounts()` is helpful in debugging what service accounts your Docker container can see.
```
jennybc, line 157: Do you think I should export `list_service_accounts()`? Because I have also used this function in my very limited GCE experimentation.

vignettes/non-interactive-auth.Rmd
```diff
@@ -122,6 +122,50 @@ options(gargle_verbosity = "debug")
...
+
+This key will now be available to add to pods within the cluster.  For Airflow, you can pass them in using the `GKEPodOperator(...., namespace='my-namespace', service_account_name='bq-service-account')`
+
+5. When calling the `gargle::gce_credentials()` within R, you need first make sure its using the right endpoint (`options(gargle.gce.use_ip = TRUE)`) and then call the service email that is not "default".  `gargle:::list_service_accounts()` is helpful in debugging what service accounts your Docker container can see. 
+
+```r
+# code within the Docker container
+
+options(gargle.gce.use_ip = TRUE)
+gargle::credentials_gce("my-service-key@my-project.iam.gserviceaccount.com")
```
jennybc, line 163: This could be a call to `PKG_auth(service_account = "my-service-key@my-project.iam.gserviceaccount.com")`, yeah? If the PKG package is using gargle in the standard way, then I think this should "just work" and is using higher-level, more user-facing functions.

jennybc (approved):
Thanks very much for this! I made a few comments inline.

Once we resolve those conversations and this is "done", I will do one round of copy edits and do stuff like "one sentence per line", FYI. And then I'll merge it.

vignettes/non-interactive-auth.Rmd
MarkEdmondson1234: It is not, its Airflow code and now I look its recently updated to GKEStartPodOperator - I will update the example with a link https://github.com/GoogleCloudPlatform/python-docs-samples/blob/HEAD/composer/workflows/gke_operator.py

vignettes/non-interactive-auth.Rmd
MarkEdmondson1234: I do think its useful and worth exporting

vignettes/non-interactive-auth.Rmd
MarkEdmondson1234: Yes, I'm stuck a bit with my "old" way to be sure which auth method is used, but the preferred way you specify should be encouraged

jennybc:
@MarkEdmondson1234 So do you consider this done and I am free to make more cosmetic / stylistic changes and merge?

MarkEdmondson1234:
Yes please do

PR #237, opened 2023-02-21
Accomodate Google Colab

### LINKED ISSUE #140: Auth in Google Colaboratory notebooks
Manually transferring from https://github.com/tidyverse/googledrive/issues/284, since I can't truly do a cross-org transfer.

This is the one about auth in Colaboratory / Jupypter notebooks, where you can't really do normal oob auth, but because `interactive()` returns `FALSE`, the normal oauth dance doesn't work either.

See also https://github.com/r-lib/httr/pull/634 for an experiment around that (the `interactive()` issue).

jennybc:
Closes #140


R/gargle-package.R
```diff
@@ -92,3 +92,25 @@ gargle_oob_default <- function() {
...
+
+#' @rdname gargle_options
+#' @export
+#' @section `gargle_oauth_client_type`:
+#' `gargle_oauth_client_type()` returns the option named
+#' "gargle_oauth_client_type", if defined. If defined, the option must be either
+#' "installed" or "web". If the option is not defined, the function returns:
+#' * "web" on RStudio Server, Posit Workbench, Posit Cloud, or Google
+#'   Colaboratory
+#' * "installed" otherwise
```
jennybc, line 104: Why is this rigged differently from `gargle_oob_default()`, where we impose our own logic *before* consulting an option? Here we first consult the option and impose our own logic if the option is undefined.

Because it seems like we must do some form of OOB auth on RStudio Server, Posit Workbench, Posit Cloud, or Google Colab. Starting a local webserver will never make sense in this context.

Whereas, once we know we're doing some form of OOB, the client type determines whether it's conventional OOB ("installed" client type) or pseudo-OOB ("web" client type). I expect conventional OOB to be relatively rare these days, but it is still possible.

R/zzz.R
```diff
@@ -1,4 +1,7 @@
 .onLoad <- function(lib, pkg) { # nocov start
+  # we need a way to tell httr that a Google Colab session is interactive
+  utils::assignInNamespace("is_interactive", rlang::is_interactive, ns = "httr")
+
```
jennybc, line 4: This can go away if https://github.com/r-lib/httr/pull/734 is merged and released.

PR #239, opened 2023-03-05
Optionally skip tests requiring mockr

MichaelChirico:


jennybc:
As of 1c06c41e7a8473320287f6f309bb32d4376aa5b4, gargle no longer uses mockr and instead uses new mocking functionality in testthat.

PR #242, opened 2023-04-03
Introduce `cli_menu()` and friends

jennybc:
This is functionally important for gargle, because `cli_menu()` is ultimately implemented on top of `readline()`, not `utils::menu()`. And Jupyter has shimmed `readline()` to work even though, at the time of writing, `base::interactive()` returns `FALSE` in Jupyter.

Therefore this PR makes gargle's interactive choices work in Jupyter notebooks, which notably covers Google Colab.

This is part (but not all) of making user auth "just work" on Colab (#140).

I have adapted `cli_menu()` from https://github.com/rstudio/rsconnect/blob/main/R/utils-cli.R. I think most of my changes might be generally positive and could be considered when/if something like this goes into cli itself (https://github.com/r-lib/cli/issues/228). I also added tests.

@hadley @gaborcsardi I don't really _need_ feedback on this (though it's welcome). This is more of an FYI request for review, to continue the discussion around https://github.com/r-lib/cli/issues/228. I'll make a few comments inline.

---

Links to the Jupyter + `readline()` story:

https://github.com/IRkernel/IRkernel/pull/452

https://github.com/IRkernel/IRkernel/blob/1eddb304b246c14b62949abd946e8d4ca5080d25/R/execution.r#L131-L137

https://github.com/IRkernel/IRkernel/blob/1eddb304b246c14b62949abd946e8d4ca5080d25/R/execution.r#L271-272


R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+
+# menu(), but based on readline() + cli and mockable ---------------------------
+# https://github.com/r-lib/cli/issues/228
+# https://github.com/rstudio/rsconnect/blob/main/R/utils-cli.R
+
+cli_menu <- function(header,
+                     prompt,
+                     choices,
+                     not_interactive = choices,
+                     exit = integer(),
```
jennybc, line 283: I called this `exit` (instead of `quit`) to better match the typical vocabulary here (e.g. in the docs for `utils::menu()`) and to avoid any confusion with actuallly quitting R.

R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+  }
+
+  choices <- paste0(cli::style_bold(seq_along(choices)), ": ", choices)
+  cli::cli_inform(
+    c(header, prompt, choices),
+    .envir = .envir
+  )
+
+  # guard against invalid mocked input
+  local_input <- getOption("cli_input", character())
```
jennybc, line 301: I was able to get into an infinite loop without this. Of course it only happens when one enters invalid mocked input, but still seemed worth accounting for.

R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+    if (selected %in% c("0", seq_along(choices))) {
+      break
+    }
+    if (length(local_input) > 0) {
+      cli::cli_abort(
+        c(x = "Internal error: mocked input is invalid."),
+        .envir = .envir,
+        call = error_call
+      )
+    }
```
jennybc, line 314: The second part of "avoiding an infinite loop".

R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+    cli::cli_inform(
+      "Enter a number between 1 and {length(choices)}, or enter 0 to exit."
+    )
+  }
+
+  selected <- as.integer(selected)
+  if (selected %in% c(0, exit)) {
+    if (is_testing()) {
+      cli::cli_abort("Exiting...", call = NULL)
```
jennybc, line 323: Exiting not quitting

R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+    } else {
+      cli::cli_alert_danger("Exiting...")
```
jennybc, line 325: Ditto

R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+      # simulate user pressing Ctrl + C
+      invokeRestart("abort")
```
jennybc, line 327: Omitted the rogue `cnd` that the rsconnect version has here (presumably a think-o).

R/utils-ui.R
```diff
@@ -271,3 +271,89 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+    input <- local_input[[1]]
+    cli::cli_inform(paste0(prompt, input))
+    options(local_input = local_input[-1])
+    input
+  } else {
+    readline(prompt)
+  }
+}
+
+local_user_input <- function(x, env = caller_env()) {
```
jennybc, line 349: Went with a `local_*()`-style name

tests/testthat/test-utils-ui.R
```diff
@@ -58,3 +58,94 @@ test_that("bulletize() works", {
   expect_snapshot(cli::cli_bullets(bulletize(letters[1:6], n_fudge = 0)))
   expect_snapshot(cli::cli_bullets(bulletize(letters[1:8], n_fudge = 3)))
 })
+
+# menu(), but based on readline() + cli and mockable ---------------------------
```
jennybc, line 62: This basically covers the `cli_menu()` stuff 100%, except the lines you can only hit in interactive use.

gaborcsardi (approved):
👍

R/Gargle-class.R
```diff
@@ -273,14 +273,15 @@ encourage_httpuv <- function() {
...
-     in many cases.",
-    "It doesn't seem to be installed.",
-    "Would you like to install it now?"
-  ))
-  if (utils::menu(c("Yes", "No")) == 1) {
+  choice <- cli_menu(
+    c(
+      "The {.pkg httpuv} package enables a nicer Google auth experience, in many cases.",
+      "It doesn't seem to be installed."
+    ),
```
hadley, line 280:
```suggestion
   "The {.pkg httpuv} package enables a nicer Google auth experience, in many cases, but it isn't installed.",
```
?

R/oauth-cache.R
```diff
@@ -315,13 +317,14 @@ token_match <- function(candidate, existing, package = "gargle") {
...
-  local_gargle_verbosity("info")
-  gargle_info(c(
+  choice <- cli_menu(
     "The {.pkg {package}} package is requesting access to your Google account.",
-    "Select a pre-authorised account or enter '0' to obtain a new token.",
-    "Press Esc/Ctrl + C to cancel."
-  ))
-  choice <- utils::menu(emails)
+    c(
+      "Select a pre-authorised account or enter '0' to obtain a new token.",
```
hadley, line 323: Should we add an explicit bullet like "Obtain a new token" instead of using 0? (I don't think this code will work now because `cli_menu()` will exit to the top-level if the user enters 0)

R/utils-ui.R
hadley: Can we do that all in one place? Or extract it out into a helper?

hadley (approved):
Looks like some nice improvements!

R/utils-ui.R
jennybc: OK I consolidated things, just to make it a bit neater, at least.

R/oauth-cache.R
jennybc: > I don't think this code will work now because `cli_menu()` will exit to the top-level if the user enters 0

Yes, this is true. I will deal with that :thinking:

R/utils-ui.R
hadley: Why does it error if it’s length greater than zero? The original intent was to be able to supply multiple values that were used in turn.

R/oauth-cache.R
```diff
@@ -316,20 +316,23 @@ token_match <- function(candidate, existing, package = "gargle") {
...
+    "Send me to the browser for a new auth process.",
+    extract_email(existing)
+  )
   choice <- cli_menu(
     "The {.pkg {package}} package is requesting access to your Google account.",
     c(
-      "Select a pre-authorised account or enter '0' to obtain a new token.",
-      "Press Esc/Ctrl + C to cancel."
+      "Enter '1' to start a new auth process or select a pre-authorized account.",
+      "Press Esc or Ctrl + C to cancel."
```
hadley, line 327: Maybe you could omit this line now that `0` has its typical behaviour?

R/utils-ui.R
jennybc: Over in rsconnect, if I add this test, it causes the test suite to just hang:

```
test_that("cli_menu() works with multiple inputs", {
  simulate_user_input(3)

  expect_snapshot(
    cli_menu(
      "Let's talk",
      "Are you OK?",
      choices = c("Yes", "No")
    )
  )
})
```
Screenshot 2023-04-04 at 9 24 46 PM

I guess I haven't found an organic need to supply multiple values. But I _have_ managed to supply an invalid one. So it seems nice for development purposes to handle that case.

I can try to think about how to preserve both.

jennybc: Are you convinced that multiple mocked inputs is important?

It feels like we could just make multiple calls to `simulate_user_input()` (rsconnect) or `local_user_input()` (here in gargle) in such a test and let the functions be simpler.

jennybc: I realize I can pursue the simplification that's most natural here. I'm mostly talking about what we might eventually want in cli or in some standalone file for general use.

R/utils-ui.R
```diff
@@ -250,3 +250,80 @@ compute_n_show <- function(n, n_show_nominal = 5, n_fudge = 2) {
...
+  withr::local_options(
+    rlang_interactive = TRUE,
+    # trailing 0 prevents infinite loop if x only contains invalid choices
+    cli_input = c(x, "0"),
```
jennybc, line 322: Appending `"0"` here means you can't get stuck in an infinite loop, which means I can also delete the fussy check inside the `repeat { ... }` in `cli_menu()`.

Maybe this makes both of us happy? 😅

jennybc:
I found a solution I'm like to the infinite loop problem, which is to append `"0"` to any mocked user input. And this retains the ability to provide multiple inputs.

R/utils-ui.R
jennybc: Yeah, that's an interesting idea. I've mentally moved on to other projects in gargle, but I think this conversation and the state of these functions here can still inform whatever a more official version looks like.

PR #247, opened 2023-04-10
Refactor GCE flow

### LINKED ISSUE #216: Oauth scopes are being ignored during credentials_gce() auth flow
Hi,
we're having problems setting scopes in credentials_gce() because the scopes are being ignored at [credentials_gce.R](https://github.com/r-lib/gargle/blame/8749ceaba896253ac21648f05f6798285205b2a3/R/credentials_gce.R#L146).

Our RStudio connect pod runs with a different GKE service account that's binded to a separate IAM service account. This service account is different from the one that's attached to our node pool. We tried adding default oauth scopes to the node pool, but that didn't have any effect.

We would like to have the ability to set the scopes with gargle, because other solutions include downloading the service account and attaching it to the pod which are not secure.

My proposed solution is to add the scopes in fetch_gce_access_token function:
```r
fetch_gce_access_token <- function(scopes, service_account) {
  path <- glue("instance/service-accounts/{service_account}/token")
  if(!is.null(scopes)) {
    scopes <- paste(scopes, collapse = ",")
    path <- glue("{path}?scopes={scopes}")
  }
  response <- gce_metadata_request(path)
  httr::content(response, as = "parsed", type = "application/json")
}
```

jennybc:
I'm refactoring the GCE flow. `credentials_gce()` is sort of the last semi-untouched frontier for me, within gargle. I have made some changes here before, but with a very light hand.

My cumulative experience with gargle and even with the GCE flow has now reached a point where it feels right to refactor this to be more like the other flows and to be more natural to me.

Issue-wise, #216 is the most relevant. Closes #216.


R/credentials_gce.R
```diff
@@ -74,36 +74,60 @@
...
+  missing <- setdiff(requested_scopes, account_scopes)
+  if (length(missing) > 0) {
+    gargle_debug(c(
+      "!" = "{cli::qty(length(missing))}{?This/These} requested \\
+             scope{?s} {?is/are} not among the scopes for the \\
+             {.val {service_account}} service account:",
+      bulletize(missing, bullet = "x"),
+      "i" = "If there are problems downstream, this might be the root cause."
+    ))
+  }
```
jennybc, line 98: This whole scope checking block is new.

Pragmatically, I could do this comparison (instance scopes vs. requested scopes) only when the `"gargle_verbosity"` option is `"debug"`. But it feels like a better practice to use the option in the narrowest sense, i.e. only to control messaging.

R/credentials_gce.R
```diff
@@ -74,36 +74,60 @@
...
-  token <- GceToken$new(
-    credentials = gce_token$access_token,
-    params = params,
-    cache_path = FALSE,
-    # The underlying Token2 class appears to *require* an endpoint and an app,
-    # though it doesn't use them for anything in this case.
-    endpoint = httr::oauth_endpoints("google"),
-    app = httr::oauth_app("google", key = "KEY", secret = "SECRET")
-  )
-  token$refresh()
```
jennybc, line 98: This logic has been relocated as I refactor this to look more like all the other token flows.

Various simplifications are also happening at the same time:

* I no longer need to set `cache_path = FALSE` or to to set the `endpoint` and `app` in light of the broader changes to the `GceToken` class.
* I don't call `token$refresh()` here anymore. Let's talk about that!

The original implementation basically fetched a token (token one) then immediately refreshed that token, which redoes the whole flow (token two). I _think_ that is not necessary? But would love to hear if I am wrong.

I think, if the current access token is found to have expired, downstream requests will see that this token is refreshable and will refresh (i.e. start over) as needed.

R/credentials_gce.R
```diff
@@ -74,36 +74,60 @@
...
   if (is.null(token$credentials$access_token) ||
     !nzchar(token$credentials$access_token)) {
     NULL
   } else {
+    gargle_debug("GCE service account email: {.email {service_account_details$email}}")
+    gargle_debug("GCE service account name: {.val {token$params$service_account}}")
+    gargle_debug("GCE access token scopes: {.val {commapse(base_scope(token$params$scope))}}")
```
jennybc, line 108: Just emitting more easy-to-get specifics for transparency.

R/credentials_gce.R
```diff
@@ -74,36 +74,60 @@
...
   }
 }
 
+#' Fetch access token for a service account on GCE
+#'
+#' @inheritParams credentials_gce
+#'
+#' @keywords internal
+#' @export
+gce_access_token <- function(scopes = "https://www.googleapis.com/auth/cloud-platform",
```
jennybc, line 119: This sort of adapter function is common in gargle's other flows. It sits between the `credential_*()` function that is suitable for inclusion in the credential registry used by `token_fetch()` and the R6 class (`GceToken`, in this case).

R/credentials_gce.R
```diff
@@ -114,36 +138,112 @@ credentials_gce <- function(scopes = "https://www.googleapis.com/auth/cloud-plat
...
+
+    missing <- setdiff(self$params$scope, actual_scopes)
+    if (length(missing) > 0) {
+      gargle_debug(c(
+        "!" = "{cli::qty(length(missing))}{?This/These} requested \\
+             scope{?s} {?is/are} not among the scopes for the \\
+             access token returned by the metadata server:",
+        bulletize(missing, bullet = "x"),
+        "i" = "If there are problems downstream, this might be the root cause."
+      ))
```
jennybc, line 177: Another scope comparison block, this time comparing the token's *actual* scopes to those that were requested. It looks a lot like the previous one (comparing the instance scopes to requested scopes), but I've only copy/pasted twice, so no helper yet. I predict I'll revisit the GCE flow as we learn more from this extra info and I can tighten things up then.

R/credentials_gce.R
```diff
@@ -114,36 +138,112 @@ credentials_gce <- function(scopes = "https://www.googleapis.com/auth/cloud-plat
...
+      gargle_debug(c(
+        "!" = "Setting token scopes to:",
+        bulletize(actual_scopes)
+      ))
+      self$params$scope <-actual_scopes
```
jennybc, line 182: This is new-ish, in the sense that I'm recording the token's actual scopes vs. recording the requested scopes. Of course, the actual scopes were always whatever they were, even if it's not what we asked for. With this change, we're just being more honest with ourselves about what the token we just got can actually do. This will affect a `$refresh()` call, i.e. these will be the requested scopes upon refresh.

R/credentials_gce.R
```diff
@@ -114,36 +138,112 @@ credentials_gce <- function(scopes = "https://www.googleapis.com/auth/cloud-plat
...
+  },
+  #' @description Placeholder implementation of required method. Returns `TRUE`.
   can_refresh = function() {
     TRUE
   },
-  #' @description Refresh a GCE token
-  refresh = function() {
-    # The access_token can only include the token itself, not the expiration and
-    # type. Otherwise, the httr code will create extra header lines that bust
-    # the POST request:
```
jennybc, line 137: A bit of fussiness was introduced in #98 that I now think was a red herring. I merged it, but at the time I had very little understanding of this flow and definitely had no hands-on experience with it (whereas I do now).

I'm getting rid of this fussiness now.

R/credentials_gce.R
```diff
@@ -114,36 +138,112 @@ credentials_gce <- function(scopes = "https://www.googleapis.com/auth/cloud-plat
...
-    gce_token <- fetch_gce_access_token(
-      self$params$scope,
-      service_account = self$params$service_account
+
+  #' @description Format a [GceToken()].
+  #' @param ... Not used.
+  format = function(...) {
```
jennybc, line 212: Level up the print method to the beautiful standard of the other flows 💅

R/credentials_gce.R
```diff
@@ -235,7 +335,8 @@ gce_instance_service_accounts <- function() {
 # https://github.com/r-lib/gargle/issues/216
 fetch_gce_access_token <- function(scopes, service_account) {
   path <- glue("computeMetadata/v1/instance/service-accounts/{service_account}/token")
-  response <- gce_metadata_request(path)
+  scope_string <- glue_collapse(scopes, sep = ",")
+  response <- gce_metadata_request(path, query = list(scopes = scope_string))
```
jennybc, line 339: This is actually sort of a big change: we now forward the requested scopes to this request for the access token.

This is in response to #216. In my hands, I still can't really see the effect of this, in the sense that my tokens always reflect the instance scopes and not what's requested here. But it also doesn't seem to hurt anything and I can imagine that my super simple GCE usage doesn't cover all of the use cases.

jennybc (commented):
I've made some inline comments, 50% for myself and 50% for any reviewer who comes by.

I don't expect anyone else to have this uploaded in their head, but all of you folks I'm tagging as reviewer know something about this and I would happy to hear any observations you have. Think of this more as an invitation to comment than a cry for help.

R/credentials_gce.R
jennybc: What this actually looks like in the case of `googledrive::drive_find()`:

```
GceToken initialize
GceToken init_credentials
! This requested scope is not among the scopes for the access token returned by the metadata server:
✖ https://www.googleapis.com/auth/drive
ℹ If there are problems downstream, this might be the root cause.
```

R/credentials_gce.R
jennybc: What this looks like in the debugging output:

```
! Setting token scopes to:
• https://www.googleapis.com/auth/cloud-platform
```

jennybc:
@atheriel I can't tag you as a reviewer, but I would welcome your 👀 on this.

R/credentials_gce.R
```diff
@@ -74,36 +74,61 @@
...
                             service_account = "default", ...) {
-  gargle_debug("trying {.fun credentials_gce}")
+  gargle_debug("Trying {.fun credentials_gce} ...")
   if (!is_gce()) {
+    gargle_debug(c("x" = "We don't seem to be on GCE."))
     return(NULL)
   }
 
-  gce_token <- fetch_gce_access_token(scopes, service_account = service_account)
+  scopes <- scopes %||% "https://www.googleapis.com/auth/cloud-platform"
```
jennybc, line 83: `token_fetch()` sets `scopes = NULL` by default, so this just reinforces the strongly preferred default in the GCE setting.

R/credentials_gce.R
```diff
@@ -74,36 +74,60 @@
 #' }
 credentials_gce <- function(scopes = "https://www.googleapis.com/auth/cloud-platform",
                             service_account = "default", ...) {
-  gargle_debug("trying {.fun credentials_gce}")
+  gargle_debug("Trying {.fun credentials_gce} ...")
```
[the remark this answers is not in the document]
hadley: Is there a function that turns debug mode on, or do you expect users to set an option?

R/credentials_gce.R
```diff
@@ -74,36 +74,61 @@
...
   if (!is_gce()) {
+    gargle_debug(c("x" = "We don't seem to be on GCE."))
     return(NULL)
   }
 
-  gce_token <- fetch_gce_access_token(scopes, service_account = service_account)
+  scopes <- scopes %||% "https://www.googleapis.com/auth/cloud-platform"
+  requested_scopes <- normalize_scopes(scopes)
+  dat <- gce_instance_service_accounts()
+  service_account_details <- as.list(dat[dat$name == service_account, ])
```
hadley, line 86: I want to say `drop = FALSE` but it probably doesn't matter.

hadley (approved):
Probably not very useful, but I took a quick pass through.

R/credentials_gce.R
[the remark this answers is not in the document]
atheriel: I've been using this quite a bit to debug auth flows:
```
gargle::with_gargle_verbosity("debug", gargle::token_fetch())
```

atheriel:
My understanding of the service account story on GCE is not strong enough to spot actual issues with the implementation, but overall the changes make sense to me and seem consistent with what I've seen in the external account flow.

PR #255, opened 2023-05-16
General modernization of some errors; better handling of HTML error

### LINKED ISSUE #254: Could not parse cli `{}` expression
In the last couple of days I have started to see some transient errors upon uploading a file to Google Drive via the googledrive package. The errors seem to be originating in the gargle package, which is why I am posting this here. Unfortunately, I cannot provide a reproducible example since the error doesn't happen every time. I am unable to identify any pattern in when it fails or doesn't. Below is the error and traceback. Any advice would be appreciated.

```
Error in `fun(..., .envir = .envir)`:
! Could not parse cli `{}` expression: ` font-family: ver...`.
Caused by error in `parse(text = code, keep.source = FALSE)`:
! <text>:1:22: unexpected ','
1:  font-family: verdana,
                         ^
Type .Last.error to see the more details.
```

```
Backtrace:
 1. global upload_drive_file("C:\\temp.7z", "Data")
 2. googledrive::drive_upload(file_path, path = id, overwrite = overwrite) at google_drive.R:89:5
 3. googledrive::as_dribble(list(gargle::response_process(response)))
 4. gargle::response_process(response)
 5. gargle:::gargle_abort_request_failed(error_message(resp), resp)
 6. gargle:::gargle_abort(message, class = c("gargle_error_request_failed", ...
 7. cli::cli_abort(message, class = c(class, "gargle_error"), .envir = .envir, ...
 8. cli:::vcapply(message, format_inline, .envir = .envir)
 9. base::vapply(X, FUN, FUN.VALUE = character(1), ..., USE.NAMES = USE.NAMES)
10. local error_message(resp)
11. gargle::response_as_json(resp)
12. gargle:::check_for_json(resp)
13. gargle:::gargle_abort_request_failed(c(gargle_map_cli(type, template = "Expected content type {.field application/json}, not \\\n                    {.field <<x>>}."), ...
14. gargle:::gargle_abort(message, class = c("gargle_error_request_failed", ...
15. cli::cli_abort(message, class = c(class, "gargle_error"), .envir = .envir, ...
16. cli:::vcapply(message, format_inline, .envir = .envir)
17. base::vapply(X, FUN, FUN.VALUE = character(1), ..., USE.NAMES = USE.NAMES)
18. local FUN(X[[i]], ...)
19. cli::cli_fmt(fun(..., .envir = .envir), collapse = collapse, strip_newline = TRUE)
20. cli:::cli__rec(expr)
21. local fun(..., .envir = .envir)
22. cli:::cli__message("inline_text", list(text = glue_cmd(..., .envir = .envir, ...
23. "id" %in% names(args)
24. cli:::glue_cmd(..., .envir = .envir, .call = sys.call(), .trim = FALSE)
25. cli:::glue(str, .envir = .envir, .transformer = transformer, .cli = TRUE, ...
26. (function (expr) ...
27. .transformer(expr, .envir) %||% character()
28. local .transformer(expr, .envir)
29. parse(text = code, keep.source = FALSE) %??% cli_error(call. = caller, ...
30. cli:::chain_error(expr, err, srcref = utils::getSrcref(sys.call()))
31. | base::withCallingHandlers({ ...
32. base::parse(text = code, keep.source = FALSE)
33. base::.handleSimpleError(function (e) ...
34. | local h(simpleError(msg, call))
35. | cli:::throw_error(err, parent = e)
```

jennybc:
Start to use more modern features of our error-throwing tools. Do a better job of surfacing error content that comes as HTML, which is gross, but happens.

Fixes #254

(This only deals with errors in the response processing domain. Wiring `call` into the auth part of gargle would be harder and have less payoff, since the vast majority of usage happens via `token_fetch()` which silently swallows errors. So it's not clear if I will ever do that.)


R/response_process.R
```diff
@@ -107,40 +113,38 @@ check_for_json <- function(resp, call = caller_env()) {
...
 
-  content <- httr::content(resp, as = "text")
   gargle_abort_request_failed(
-    c(
-      gargle_map_cli(
-        type,
-        template = "Expected content type {.field application/json}, not \\
-                    {.field <<x>>}."
-      ),
-      "*" = obfuscate(content, first = 197, last = 0)
```
jennybc, line 118: Note: this is where the `Could not parse cli `{}` expression: ...` error was coming from (#254). When the `content` is html, it can contain curly braces (and, indeed, in this case, did), which then confuses cli when it processes the error message bullets.

There is a solution, which is to refer to the `{`-and-`}`-containing string from a template:

https://cli.r-lib.org/reference/inline-markup.html#escaping-and-

However, I decided not to do that, but instead to make it much easier to see the full html. Previously, my only direct experience of html errors was from the token info endpoints, which seemed rather obscure and the examples I knew about produced a relatively small amount of html.

But the error seen in #254 is actually important for the user (429 too many requests, from googledrive) and involved considerably more html (so a wee little excerpt didn't reveal anything useful).

R/response_process.R
```diff
@@ -107,40 +113,38 @@ check_for_json <- function(resp, call = caller_env()) {
...
 }
 
 #' @export
 #' @rdname response_process
-gargle_error_message <- function(resp) {
-  content <- response_as_json(resp)
+gargle_error_message <- function(resp, call = caller_env()) {
+  type <- httr::http_type(resp)
+  if (type == "text/html") {
+    return(gargle_html_error_message(resp))
```
jennybc, line 144: Instead of letting html trigger an error in `check_for_json()` inside `response_as_json()`, I handle that case explicitly here.

tests/testthat/_snaps/response_process.md
```diff
@@ -1,86 +1,124 @@
...
-    Expected content type 'application/json', not 'text/html'.
-    * Not Found
+    Code
+      response_process(resp)
+    Condition
+      Error in `expect_recorded_error()`:
+      ! Client error: (404) Not Found
+      x Expected content type 'application/json', not 'text/html'.
+      i See 'VOLATILE_FILE_PATH' for the html error content.
+      i Or execute `browseURL("VOLATILE_FILE_PATH")` to view it in your browser.
```
jennybc, line 123: This is a look at the new handling for an error that comes as html. In certain contexts, such as RStudio, the "See 'VOLATILE_FILE_PATH' ..." is a clickable link to open the file. I can't make the `browseURL()` a clickable link because it calls into a base package. But the user can copy it, paste into the console, and execute without too much trouble.

https://cli.r-lib.org/reference/links.html#security-considerations

R/response_process.R
jennybc: This pitfall is also exactly described here:

https://rlang.r-lib.org/reference/topic-condition-formatting.html#transitioning-from-abort-to-cli-abort-

Issue #283, opened 2024-04-01
gargle depends on lifecycle v0.2.0 or higher but does not specify minimum version


pettyalex:
Hello,

Gargle seems to need a function that was added to lifecycle in v0.2.0: https://github.com/r-lib/lifecycle/releases/tag/v0.2.0 , but gargle does not specify a minimum lifecycle version.

My group only happened to encounter this because lifecycle v0.1.0 is what's in the Ubuntu 20.04 repositories bundled with `r-cran-tidyverse`: https://packages.ubuntu.com/focal/amd64/r-cran-lifecycle

This could be easily fixed by adding `>= 2.0.0` to lifecycle in DESCRIPTION.

```r
library(devtools)
install_version('lifecycle', version='0.1.0')
install.packages('gargle')
```

```r
Error: package or namespace load failed for ‘gargle’:
 .onLoad failed in loadNamespace() for 'gargle', details:
  call: NULL
  error: 'is_present' is not an exported object from 'namespace:lifecycle'
Error: loading failed
Execution halted
```

PR #284, opened 2024-04-01
Add a necessary minimum version to lifecycle

pettyalex:
Fixes https://github.com/r-lib/gargle/issues/283.

This is a niche failure, but also a very easy fix.


jennybc:
Thanks! Sorry it took so long.

